PR #2, opened 2018-06-10
Add vfs bindings

Shelnutt2:
This add VFS, Context, Config and Version bindings. Travis ci is configure (although untested at the moment).

I will add new issues for the remaining functionality needed (Array/Query/KV, etc)


jakebolewski:
I've made this repo public so that we can get CI working.

PR #8, opened 2018-06-16
Add Array Support

### LINKED ISSUE #7: Add Support For Array
Add support for Array Schema:

- [x] Attribute support
- [x] Dimension support
- [x] Domain support
- [x] ArraySchema support
- [x] Group Support
- [x] Array Support

Shelnutt2:
This add Array Support, Array Schema, Dimension, Domain, Attribute, Group and adjust VFS to store context on struct.

Test coverage is only about 58% as query support is not implemented so any code paths that require data/non-empty arrays can't be tested via unit testing yet.

For functions which take void pointers to handle arrays of varies datatypes I've handled that by taking interface{}'s as a parameter to the function and then switching on the domain/dimension datatype to be able to convert the golang array to a void * in C.

Closes #7.

### DIFF
--- a/dimension.go
+++ b/dimension.go
@@ -1,0 +1,381 @@
+package tiledb
+
+/*
+#cgo LDFLAGS: -ltiledb
+#include <tiledb/tiledb.h>
+#include <stdlib.h>
+*/
+import "C"
+
+import (
+	"fmt"
+	"reflect"
+	"runtime"
+	"strconv"
+	"unsafe"
+)
+
+// Dimension is tiledb dimension
+type Dimension struct {
+	tiledbDimension *C.tiledb_dimension_t
+	context         *Context
+}
+
+// NewDimension alloc a new dimension
+func NewDimension(context *Context, name string, domain interface{}, extent interface{}) (*Dimension, error) {
+	dimension := Dimension{context: context}
+	var cname *C.char = C.CString(name)
+	defer C.free(unsafe.Pointer(cname))
+
+	if reflect.TypeOf(domain).Kind() != reflect.Slice {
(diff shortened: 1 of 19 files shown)


dimension.go
```diff
@@ -0,0 +1,381 @@
...
+	context         *Context
+}
+
+// NewDimension alloc a new dimension
+func NewDimension(context *Context, name string, domain interface{}, extent interface{}) (*Dimension, error) {
+	dimension := Dimension{context: context}
+	var cname *C.char = C.CString(name)
+	defer C.free(unsafe.Pointer(cname))
+
+	if reflect.TypeOf(domain).Kind() != reflect.Slice {
```
jakebolewski: Is there a way here to check if the domain slice type == extent slice type?  Probably best to enforce this as issues with auto type conversion can be tricky to debug.

jakebolewski:
Few minor comments, looks great!

Shelnutt2:
I have adjusted the LoadArraySchema to be a function that is not associated to the struct. There are no static functions in golang for being member of a struct. Having this create and read the function will solve other issues of having partial structs due to bad reads, and having the already initialized struct. 

I realized there were a few important functions missing from the ArraySchema class, I added those also.

The config's LoadFromFile function is also a struct member, I see in the c++ api the construct just takes an optional path for loading from file. Unless you object I'll make the same change to the `NewConfig()` function so it takes an optional path and loads config from disk that way.

jakebolewski:
👍 sounds good to me

dimension.go
Shelnutt2: A safety check for this is simple to add, I'll do that next.

PR #10, opened 2018-06-27
Query operation support

### LINKED ISSUE #9: Add support for Query operations
Add support for Query operations by implementing a query struct similar to c++ api.

Shelnutt2:
This add support for querying. A full example is given for the docs for NewQuery showing how to create and use an array with fixed and variable attributes.

Current limitations:
- SubmitAsync does not support a call back function. Support could be added for passing a void* in but this will require the end user to write a cgo function themselves. A good example is in the [cgo wiki article](https://github.com/golang/go/wiki/cgo#function-pointer-callbacks). I believe we'd need to document this and provide examples for users, so for now I've left off support.
  - An alternative to use the callback from the c api is we could start a goroutine that polls the status and calls a user pass function when status is complete. This is simplier to an end user, but it would be nice to keep the functionality in the c api instead of reproducing it.
- Encoding/Decoding utf-16/utf-32/ucs-2/ucs-4 is left to a user, the SetBuffers will expect only uint16/32 array. At somepoint it'd be nice to handle the encoding/decoding for the user
- Same goes for strings (utf-8/ascii), a []byte (uint8) array is expected, it'd be nice to convert to/from go string for a user if they pass that as the buffer.

Closes #9

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ go test github.com/TileDB-Inc/TileDB-Go
 
 ## Example Usage
 
-Below is a small example using vfs functionality. Additionale examples are
+Below is a small example using vfs functionality. Additional examples are
 provided in the GoDoc documentation.
 
 ```golang
--- a/array.go
+++ b/array.go
@@ -38,7 +38,7 @@ func NewArray(ctx *Context, uri string) (*Array, error) {
 	return &array, nil
 }
 
-// Free tiledb_domain_t that was allocated on heap in c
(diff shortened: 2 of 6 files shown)


Shelnutt2:
@jakebolewski I've adjusted the code to now store the buffer slices on an slice on the query struct. I also realized we did not need to take pointers for the buffer since slices are just a header struct with a pointer to the array.

I modified the write unit test to nullify of the buffer right after called SetBufferVar to confirm the underlying array isn't gc'ed before written.

PR #13, opened 2018-06-29
Add tiledb stats bindings

### LINKED ISSUE #12: Add support for Stats api
Add support for Stats api

Shelnutt2:
Add tiledb stats bindings including dump support to a given filepath and to stdout.

Closes #12

### DIFF
--- a/stats.go
+++ b/stats.go
@@ -1,0 +1,78 @@
+package tiledb
+
+/*
+#cgo LDFLAGS: -ltiledb
+#include <tiledb/tiledb.h>
+#include <stdio.h>
+#include <stdlib.h>
+*/
+import "C"
+
+import (
+	"fmt"
+	"os"
+	"unsafe"
+)
+
+// StatsEnable enable internal statistics gathering
+func StatsEnable() error {
+	ret := C.tiledb_stats_enable()
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error enabling stats")
+	}
+	return nil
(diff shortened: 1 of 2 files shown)


stats.go
```diff
@@ -0,0 +1,77 @@
...
+	// Convert to char *
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	// Set mode as char*
+	cMode := C.CString("w")
+	defer C.free(unsafe.Pointer(cMode))
+
+	// Open file to get FILE*
+	cFile := C.fopen(cPath, cMode)
```
jakebolewski: Does the file need be closed here?

jakebolewski: I think there needs to be a defer close here right?

jakebolewski: also if the dump is unsuccessful / throws an error for whatever reason, we should probably try to clean up the opened file

Shelnutt2: Good catch, I've added the defer C.fclose call. When you say clean up the opened file, do you mean we should truncate/delete the file (if it exists) or do you just mean to make sure it gets closed? The defer C.fclose will close the file once the scope ends so it will close at either return.

jakebolewski: On second thought just ensuring the file is closed is fine.  I was thinking about if a file was created, but not written to this could be confusing.  However there are instances where it would not make sense to remove the unsuccessfully written file (ex a system file), so ignore that suggestion

Issue #37, opened 2018-11-11
Add support for Key-Value Map


rw:
Thanks for writing this amazing library! (Both the Go wrapper and the underlying C/C++ code.) It's exciting research and I'm glad to see it applied so well.

Is there a roadmap for KV support in Go? I'd try to implement it myself based on the docs for the C++ version, but it sounds like you have special logic to make it faster than it would be otherwise.

Thanks!

Shelnutt2:
@rw I've started working on an implementation now, hopefully it will be pushed tomorrow.

In general the key-value store is not as optimized as a standard tiledb array. As you've seen the C++ api implements some components to help with the performance. Performance of the KV is something we hope to be able to address at some point in the future.

The c++ api also wraps the KV into a map container. In golang, there is no operator overloading or functionality to implement a nice map interface like that. The initial implementation I am working on will be closer to the c api for now.

rw:
@Shelnutt2 Wow, that's a quick turnaround, thanks! Is it true that the exported C functions don't support all the KV features of the C++ library, so you have to re-implement some logic in Go?

Shelnutt2:
@rw Everything needed for base functionality exported via the tiledb c api. The C++ key-value api just wraps things up into a nice c++ idiomatic way, such as exporting it via the map interface so you can do `kv["key"]["attribute_1"] = "my_value"`. It also provider c++ iterator support, and templated handling of the key or value datatype.

I've merged the initial KV support into master branc. For the go api, its not 100% idiomatic go but for a first pass I've tried to make it not too difficult to work with. You can iterate over the KV in a for loop, and you can pass keys/value via interface method. I've got an [runable example](https://github.com/TileDB-Inc/TileDB-Go/blob/0de12aca7a6287f203a1d85c59bb994d8967a2c0/kv_test.go#L13) you can take a look at.

I will try to get a tagged release within the next couple weeks. Please let me know if you have any issue or question on this initial kv bindings.

rw:
@Shelnutt2 I ended up deciding to use the C bindings from Rust :-) Thanks though!

PR #38, opened 2018-11-14
Add KV support. #37

Shelnutt2:
This add basic KV support including for loop iterator support and a full testable example.

### DIFF
--- a/kv_test.go
+++ b/kv_test.go
@@ -1,0 +1,445 @@
+package tiledb
+
+import (
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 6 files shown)


kv_test.go
```diff
@@ -0,0 +1,438 @@
...
+		return
+	}
+	err = kvItem.SetValue("a1", int8(1))
+	if err != nil {
+		// handle error
+		return
+	}
+
+	// Add the item to the kv store
+	err = kv.AddItem(kvItem)
```
jakebolewski: I do not think you are opening the kv in write mode by default?

Shelnutt2: You are correct, I've adjust the example to properly open the kv in write mode before writing. I'm surprised I did not get a tiledb error about not having it opened in write mode?

PR #40, opened 2018-11-15
Fix TILEDB_CHAR attribute type handling

Shelnutt2:
Fix TILEDB_CHAR attribute type handling and improve readme example for string attributes.


jakebolewski:
I think you need to run go-fmt

PR #45, opened 2019-01-31
SparseArray example does not work

emi80:
Problems are:
- the write query layout cannot be `ROW_MAJOR` for SparseArray
- `array.MaxBufferElements()` should return buffer bounds for coordinates too

This pull request fixes these issues by:
- modifying the write query layout to `TILEDB_UNORDERED` in the example code
- adding logic inside the `Array.MaxBufferElements` function  to handle coordinates

### DIFF
--- a/array.go
+++ b/array.go
@@ -662,6 +662,22 @@ func (a *Array) MaxBufferElements(subarray interface{}) (map[string][2]uint64, e
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Shelnutt2:
@emi80 Thank you for the contribution! Please let us know if you run into any other issues with the go bindings.

PR #46, opened 2019-02-19
uses go modules, enables tests for quickstarts

### LINKED ISSUE #32: Add support for go modules
With golang 1.11 released we should support the new experimental go modules

antalakas:
Adds go.mod file having name:
module github.com/TileDB-Inc/TileDB-Go 
Closes #32 

With go version 1.11.x and later it is no longer needed to put the code in GOPATH. After the first go test or go build command dependencies are downloaded automatically.

Quickstarts (sparse /dense) are included in the tiledb package in order to compile (avoids cyclic reference), also adds simple assertions

### DIFF
--- a/quickstart_dense_test.go
+++ b/quickstart_dense_test.go
@@ -27,13 +27,13 @@ When run, this program will create a simple 2D dense array, write some data
 to it, and read a slice of the data back, then clean up.
 For simplicity this program does not handle errors
 */
-package tiledb_test
+package tiledb
 
 import (
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"os"
-
-	tiledb "github.com/TileDB-Inc/TileDB-Go"
+	"testing"
 )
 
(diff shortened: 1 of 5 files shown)


quickstart_dense_test.go
```diff
@@ -108,15 +108,19 @@ func readDenseArray() {
...
 	// Print out the results.
 	fmt.Println(data)
+
+	return data
 }
 
 // ExampleDenseArray shows and example creation, writing and reading of a dense
 // array
-func Example_denseArray() {
+func TestDenseArray(t *testing.T) {
```
Shelnutt2: The function name should start with `Example.`. instead of `Test..` so it shows up as an [example in the documentation](https://godoc.org/github.com/TileDB-Inc/TileDB-Go#ex-package--DenseArray). Example are suppose to print their output and that is how it is check.

See https://blog.golang.org/examples

PR #53, opened 2019-03-04
Effective buffer size

### LINKED ISSUE #51: Adjust SetBuffer/SetBufferVar to return pointer for effective buffer size
An oversight of the go api is we currently provide no way to get the effective buffer size for reads. The c api provides this functionality by updating the pointer to the `buffer_size` with the effective size after reads.

Currently we set the `cBufferSize`, but do not provide a way to return to the user. We should make `bufferSize` a uint64, then pass get a pointer to it, cast it to the C pointer type, and return this pointer to the user. We will need to store a reference to this just like we do with the actual data slices to prevent it from being gc'ed (go garbage collector does not track variables passed to CGO).

```
ret := C.tiledb_query_set_buffer(q.context.tiledbContext, q.tiledbQuery, cAttribute, cbuffer, &cbufferSize)
    if ret != C.TILEDB_OK {
        return fmt.Errorf("Error setting query buffer: %s", q.context.LastError())
    }
```

Also it would be good to add a unit test in query_test.go for incomplete queries to validate this new functionality is working as expected.

antalakas:
SetBuffer, SetBufferVar and SetCoordinates are amended to return a *uint64
This return value contains the effective buffer size
Contents of memory location is materialised after Submit() is called

Closes #51

### DIFF
--- a/query.go
+++ b/query.go
@@ -15,7 +15,7 @@ import (
 	"unsafe"
 )
 
-// Query construct and exeute read/write queries on a tiledb Array
+// Query construct and execute read/write queries on a tiledb Array
 type Query struct {
 	tiledbQuery *C.tiledb_query_t
 	array       *Array
@@ -160,17 +160,23 @@ func (q *Query) SetSubArray(subArray interface{}) error {
 
 // SetBuffer Sets the buffer for a fixed-sized attribute to a query
 // The buffer must be an initialized slice
-func (q *Query) SetBuffer(attribute string, buffer interface{}) error {
+func (q *Query) SetBuffer(attribute string, buffer interface{}) (*uint64,
+	error) {
 	bufferReflectType := reflect.TypeOf(buffer)
 	bufferReflectValue := reflect.ValueOf(buffer)
 	if bufferReflectValue.Kind() != reflect.Slice {
-		return fmt.Errorf("Buffer passed must be a slice that is pre-allocated, type passed was: %s", bufferReflectValue.Kind().String())
+		return nil, fmt.Errorf(
+			"Buffer passed must be a slice that is pre"+
+				"-allocated, type passed was: %s",
+			bufferReflectValue.Kind().String())
 	}
 
 	// Next get the attribute to validate the buffer type is the same as the attribute
 	schema, err := q.array.Schema()
 	if err != nil {
-		return fmt.Errorf("Could not get array schema for SetBuffer: %s", err)
+		return nil, fmt.Errorf(
+			"Could not get array schema for SetBuffer: %s",
+			err)
 	}
 
 	var attributeType Datatype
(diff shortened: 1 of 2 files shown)


query_test.go
```diff
@@ -409,26 +550,26 @@ func TestQueryReadEmpty(t *testing.T) {
...
 
 	bufferA2 := make([]byte, 2)
-	err = query.SetBuffer("a2", bufferA2)
+	_, err = query.SetBuffer("a2", bufferA2)
 	assert.Nil(t, err)
 
 	bufferA3 := make([]float32, 5)
 	offsetBufferA3 := make([]uint64, 3)
-	err = query.SetBufferVar("a3", offsetBufferA3, bufferA3)
+	_,err = query.SetBufferVar("a3", offsetBufferA3, bufferA3)
```
Shelnutt2: There is a space missing between `_,` and `err`. Only issue found with gofmt

query.go
```diff
@@ -218,111 +237,125 @@ func (q *Query) SetBuffer(attribute string, buffer interface{}) error {
...
 		q.buffers = append(q.buffers, tmpBuffer)
 		cbuffer = unsafe.Pointer(&(tmpBuffer)[0])
 	default:
-		return fmt.Errorf("Unrecognized buffer type passed: %s", bufferType.String())
+		return nil,
+			fmt.Errorf("Unrecognized buffer type passed: %s",
+				bufferType.String())
 	}
 
+	//fmt.Println("fix --------------------")
```
Shelnutt2: Go ahead and remove these two debug comments.

query.go
```diff
@@ -615,24 +656,38 @@ func (q *Query) SetBufferVar(attribute string, offset []uint64, buffer interface
...
 		// Store slice so underlying array is not gc'ed
 		q.buffers = append(q.buffers, tmpBuffer)
 		cbuffer = unsafe.Pointer(&(tmpBuffer)[0])
 	default:
-		return fmt.Errorf("Unrecognized buffer type passed: %s", bufferType.String())
+		return nil, fmt.Errorf("Unrecognized buffer type passed: %s",
+			bufferType.String())
 	}
 
+	//fmt.Println("var --------------------")
```
Shelnutt2: These comments can also be removed

query.go
```diff
@@ -858,6 +913,11 @@ func (q *Query) Submit() error {
 	if ret != C.TILEDB_OK {
 		return fmt.Errorf("Error submitting query: %s", q.context.LastError())
 	}
+
+	//fmt.Println("++++++++++++++++++++++++")
```
Shelnutt2: Go ahead and remove these as well

query.go
```diff
@@ -918,6 +978,7 @@ func (q *Query) HasResults() (bool, error) {
 }
 
 // SetCoordinates sets the coordinate buffer
-func (q *Query) SetCoordinates(coordinates interface{}) error {
+func (q *Query) SetCoordinates(coordinates interface{}) (*uint64, error) {
+	//fmt.Println("-> Coordinates <-")
```
Shelnutt2: Can be removed

PR #54, opened 2019-03-06
Implements ResultBufferElements

### LINKED ISSUE #52: Implement ResultBufferElements
Probably the most user friendly way to handle result buffer size for reads is to create a function for `ResultBufferElements` that returns the map of name -> {offset_elements, data_elements} just like [c++](https://docs.tiledb.io/en/stable/c-api.html#c.tiledb_query_set_buffer). We just have to store these `uint64 pointers` internally (similar to how we store a reference to the data slice so it doesn't get GC'ed).

This is a follow on issue after #51

antalakas:
Implements ResultBufferElements
SetBufferVar additionally return offsetSize
Tests are added

Closes #52

### DIFF
--- a/query.go
+++ b/query.go
@@ -17,12 +17,13 @@ import (
 
 // Query construct and execute read/write queries on a tiledb Array
 type Query struct {
-	tiledbQuery *C.tiledb_query_t
-	array       *Array
-	context     *Context
-	uri         string
-	buffers     []interface{}
-	bufferMutex sync.Mutex
+	tiledbQuery          *C.tiledb_query_t
+	array                *Array
+	context              *Context
+	uri                  string
+	buffers              []interface{}
+	bufferMutex          sync.Mutex
+	resultBufferElements map[string][2]*uint64
 }
 
 /*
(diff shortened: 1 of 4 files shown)


query.go
```diff
@@ -673,15 +680,68 @@ func (q *Query) SetBufferVar(attribute string, offset []uint64,
...
+// from a read query.
+// This is a map from the attribute name to a pair of values.
+// The first is number of elements (offsets) for var size attributes, and the
+// second is number of elements in the data buffer. For fixed sized attributes
+// (and coordinates), the first is always 0.
+func (q *Query) ResultBufferElements() (map[string][2]uint64, error) {
+	elements := make(map[string][2]uint64, 0)
+
+	// Will need the schema to infer data type size for attributes
+	schema, _ := q.array.Schema()
```
Shelnutt2: We should check for error and return if we error on getting the schema

query.go
```diff
@@ -673,15 +680,68 @@ func (q *Query) SetBufferVar(attribute string, offset []uint64,
...
+
+	for attributeName, v := range q.resultBufferElements {
+		// Skip coords
+		if attributeName == "__coords" {
```
Shelnutt2: We should return the coordinates with this function too. You can get the coordinates type (to compute elements) by checking the domain type from the array schema. Coordinate will always been fixed length.

Shelnutt2: Also there is `TILEDB_COORDS` for the coordinate name string.

query.go
antalakas: checked

query.go
antalakas: Coordinates are now included in the map returned

PR #55, opened 2019-03-07
adds go formatting to travis

### LINKED ISSUE #50: Add travis ci check for formating
We should have a check in travis ci for making sure all files are formatted. If any file is not formatted via `go fmt` the check should fail. This should be the first check that is run in CI.

antalakas:
Closes #50


Shelnutt2:
Looks like you need to mark the `./.travis.gofmt.sh` as executable. Also let's move the script under the `scripts` folder we have.

PR #71, opened 2019-05-07
Add a Query getter for just the buffer sizes.

tdenniston:
Needed for serialization.


query_test.go
```diff
@@ -406,6 +406,15 @@ func TestQueryEffectiveBufferSize(t *testing.T) {
 	_, err = query.SetCoordinates(coordsWrite)
 	assert.Nil(t, err)
 
+	// Check the buffer sizes
+	offsetSize, dataSize, err := query.BufferSizeVar("a1")
+	assert.Nil(t, err)
+	assert.Equal(t, len(a1OffWrite) * 8, int(offsetSize))
```
Shelnutt2: You can use `EqualValues` to test ignoring the datatypes, just an fyi.

tdenniston: In this case I didn't know what the right way to do `sizeof` in go was -- the buffer size is returned in bytes.

tdenniston: Actually I guess the right thing to do is return num elements instead of bytes with these getters. Will update the PR.

Shelnutt2: You can replace 8 with `unsafe.Sizeof(a1OffWrite[0])`. Since a `slice` in go is a wrapper structure (like vector), you can't find the size directly, so you still need to multiple by the length.

query.go
```diff
@@ -977,6 +977,48 @@ func (q *Query) BufferVar(attributeName string) ([]uint64, interface{}, error) {
 	return offsets, buffer, nil
 }
 
+// BufferSizeVar returns the size of the backing C buffers for the given variable-length attribute
```
Shelnutt2: Something to consider for later, is in go using raw byte size is unexpected. Go users expect to use element counts. @antalakas and I are planning to eventually cleanup the go api to remove all byte size references and return element counts everywhere we can.

tdenniston: @Shelnutt2 updated the PR to return element counts instead of bytes.

PR #76, opened 2019-05-22
Serialization updates

tdenniston:
Depends on core PR https://github.com/TileDB-Inc/TileDB/pull/1273


tdenniston:
@Shelnutt2 FYI that this PR exists -- moving to upstream TileDB for CI, and using the new C API.

PR #82, opened 2019-07-23
In calls to query.Free() remove references to user buffers

Shelnutt2:
In calls to query.Free() remove references to user buffers

### DIFF
--- a/query.go
+++ b/query.go
@@ -66,6 +66,10 @@ func NewQuery(ctx *Context, array *Array) (*Query, error) {
 
 // Free tiledb_query_t that was allocated on heap in c
 func (q *Query) Free() {
+	q.bufferMutex.Lock()
+	defer q.bufferMutex.Unlock()
+	q.buffers = nil
+	q.resultBufferElements = nil
 	if q.tiledbQuery != nil {
 		C.tiledb_query_free(&q.tiledbQuery)
 	}


joe-maley:
I assume this is related to the GC issue? Should I try this out in my test environment?

Shelnutt2:
> I assume this is related to the GC issue? Should I try this out in my test environment?

Yes, I left a comment/analysis in the other thread for continuity.

query.go
```diff
@@ -66,6 +66,10 @@ func NewQuery(ctx *Context, array *Array) (*Query, error) {
 
 // Free tiledb_query_t that was allocated on heap in c
 func (q *Query) Free() {
+	q.bufferMutex.Lock()
+	defer q.bufferMutex.Unlock()
+	q.buffers = nil
+	q.resultBufferElements = nil
 	if q.tiledbQuery != nil {
 		C.tiledb_query_free(&q.tiledbQuery)
```
joe-maley: Should we set `q.tiledbQuery` to nil here, to prevent a double-free? Same for `func (b *Buffer) Free()`

Shelnutt2: `q.tiledbQuery` is set to nil/nullptr by the `C.tiledb_query_free` function.

joe-maley: Ah, thanks!

joe-maley:
LGTM

PR #85, opened 2019-07-23
Temporarily remove running reading_incomplete.

Shelnutt2:
Due to a bug in core tiledb this example is looping infinitely. This is being disabled until the bug in core is solved. To allow for publishing an updated TileDB-Go based on TileDB 1.6.0


joe-maley:
Can you double-check this diff? I don't see any functional changes.

Shelnutt2:
> Can you double-check this diff? I don't see any functional changes.

Its golang magic :). For examples, go can perform unit test on the stdout of the example. By removing the comments for `// output: ...` it stops go from running this function as a test.

See https://blog.golang.org/examples

joe-maley:
Ah, I see. Neat! LGTM

PR #86, opened 2019-10-14
Add support for range functions in TileDB-Go

antalakas:
[ch938]

### DIFF
--- a/examples/range_test.go
+++ b/examples/range_test.go
@@ -1,0 +1,220 @@
+/**
+ * @file   range_test.go
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2019 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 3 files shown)


examples/range_test.go
```diff
@@ -0,0 +1,220 @@
...
+	if _, err := os.Stat(rangeArrayName); err == nil {
+		err = os.RemoveAll(rangeArrayName)
+		checkError(err)
+	}
+
+	// Output: Error adding query range: [TileDB::Subarray] Error: Cannot add range to dimension; Range must be in the domain the subarray is constructed from
+	// Error adding query range: [TileDB::Subarray] Error: Cannot add range to dimension; Invalid dimension index
+	// Number of ranges across dimension 0 is: 1
+	// Range start for dimension 0, range 0 is: 1
+	// Range end for dimension 0, range 0 is: 4
```
antalakas: I expected this to be 3

Shelnutt2: Ranges are inclusive, and you set `var end int32 = 3`, so I believe 3 is correct.

examples/range_test.go
```diff
@@ -0,0 +1,220 @@
+/**
+ * @file   quickstart_dense_test.go
```
Shelnutt2: File name needs to be updated

examples/range_test.go
```diff
@@ -0,0 +1,220 @@
...
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2018 TileDB, Inc.
```
Shelnutt2: copyright should be 2019

Shelnutt2:
Minor changes but looks good! Feel free to merge after the comment fixes :+1:

PR #97, opened 2019-11-20
Support array metadata

antalakas:


array.go
```diff
@@ -568,3 +570,339 @@ func (a *Array) URI() (string, error) {
...
+		datatype = TILEDB_FLOAT64
+		if isSliceValue {
+			tmpValue := value.([]float64)
+			ret = C.tiledb_array_put_metadata(a.context.tiledbContext, a.tiledbArray, ckey, C.tiledb_datatype_t(datatype), valueNum, unsafe.Pointer(&tmpValue[0]))
+		} else {
+			tmpValue := value.(float64)
+			ret = C.tiledb_array_put_metadata(a.context.tiledbContext, a.tiledbArray, ckey, C.tiledb_datatype_t(datatype), valueNum, unsafe.Pointer(&tmpValue))
+		}
+	case reflect.String:
+		datatype = TILEDB_STRING_UTF8
```
Shelnutt2: Are golang strings UTF8 or ascii?

antalakas: I had the same question, the answer was `In Go, a string is in effect a read-only slice of bytes`

This is a valid string `const sample = "\xbd\xb2\x3d\xbc\x20\xe2\x8c\x98"`
Contains non-ASCII...

In essence based on https://blog.golang.org/strings i believe that UTF8 is the safest choice

Axiomatic details:
    Go source code is always UTF-8.
    A string holds arbitrary bytes.
    A string literal, absent byte-level escapes, always holds valid UTF-8 sequences.
    Those sequences represent Unicode code points, called runes.
    No guarantee is made in Go that characters in strings are normalized.

Shelnutt2: Thanks for the analysis!

Shelnutt2:
@antalakas looks like this branch should be rebased on master.

examples/array_metadata_test.go
```diff
@@ -0,0 +1,259 @@
...
+	fmt.Printf("Value Num: %d\n", arrayMetadata.ValueNum)
+	fmt.Printf("Value: %v\n", arrayMetadata.Value.(int32))
+
+	err = array.ConsolidateMetadata(nil)
+	checkError(err)
+
+	_, err = array.GetMetadataMap()
+	checkError(err)
+
+	jsonData, err := array.MakeJSONFromMetadata()
```
Shelnutt2: `MakeJSONFromMetadata` doesn't seem to exist. Also instead of a dedicate function we should just support marshal/unmarhal of the `ArrayMetadata` type. This allows the struct to be used in high level applications easier.

antalakas: `GetMetadataMap` returns `map[string]*ArrayMetadata`

If we marshal the map, current example returns:

```
{"key1":{"Key":"key1","KeyLen":4,"Datatype":0,"ValueNum":1,"Value":25},"key2":{"Key":"key2","KeyLen":4,"Datatype":0,"ValueNum":4,"Value":[25,26,27,28]},"key3":{"Key":"key3","KeyLen":4,"Datatype":2,"ValueNum":1,"Value":25.1},"key4":{"Key":"key4","KeyLen":4,"Datatype":2,"ValueNum":4,"Value":[25.1,26.2,27.3,28.4]},"key5":{"Key":"key5","KeyLen":4,"Datatype":12,"ValueNum":29,"Value":"This is TileDb array metadata"}}
```

antalakas: `MakeJSONFromMetadata` currently returns:

```
{
   "key1":25,
   "key2":[
      25,
      26,
      27,
      28
   ],
   "key3":25.1,
   "key4":[
      25.1,
      26.2,
      27.3,
      28.4
   ],
   "key5":"This is TileDb array metadata"
}
```

PR #100, opened 2020-01-13
Handle 64bit integer for slice sizes

Shelnutt2:
Handle 64bit integer for slice sizes


Shelnutt2:
> Why do we need this now? Why we did not use 64bit at first place?

We need this now because @joe-maley was testing trying to use a 2 gigabyte buffer and it failed with out of bounds. We did not use 64bits before simply as an oversight from following an example which used 32bits.

PR #101, opened 2020-01-15
updates tests

antalakas:


antalakas:
WIP: Still missing ExampleWritingSparseMultiple

examples/writing_sparse_global_test.go
```diff
@@ -99,24 +100,12 @@ func writeSparseGlobalArray() {
...
-	checkError(err)
-	_, err = query.SetCoordinates(coords1)
-	checkError(err)
-
-	// Perform the write.
-	err = query.Submit()
-	checkError(err)
-
-	// Submit second query
-	coords2 := []int32{2, 3}
```
Shelnutt2: I think we can keep this but just create a second query object?

antalakas: Ok, will do, especially because in the case of `ExampleWritingSparseMultiple` the purpose of the test is cancelled

antalakas: Done

PR #108, opened 2020-04-10
adds string dimension support

antalakas:
ExampleStringDimArray follows https://github.com/TileDB-Inc/TileDB/blob/e26d52bc20996d9dd320032aa3049dd198c5c7fc/test/src/unit-cppapi-array.cc#L1120

### DIFF
--- a/array.go
+++ b/array.go
@@ -254,65 +254,60 @@ func (a *Array) QueryType() (QueryType, error) {
 	return QueryType(queryType), nil
 }
 
-// makeNonEmptyDomain creates a []NonEmptyDomain from a generic domain-typed slice
-func makeNonEmptyDomain(domain *Domain, domainSlice interface{}) ([]NonEmptyDomain, error) {
-	domainType, err := domain.Type()
-	if err != nil {
-		return nil, err
-	}
-	ndims, err := domain.NDim()
+// getNonEmptyDomainForDim creates a NonEmptyDomain from a generic dimension-typed slice
+func getNonEmptyDomainForDim(dimension *Dimension, dimensionSlice interface{}) (*NonEmptyDomain, error) {
+	dimensionType, err := dimension.Type()
 	if err != nil {
 		return nil, err
 	}
 
-	nonEmptyDomains := make([]NonEmptyDomain, 0)
-	for i := uint(0); i < ndims; i++ {
-		dimension, err := domain.DimensionFromIndex(i)
-		if err != nil {
-			return nil, err
-		}
-		name, err := dimension.Name()
-		if err != nil {
-			return nil, err
-		}
-
-		switch domainType {
-		case TILEDB_INT8:
-			tmpDomain := domainSlice.([]int8)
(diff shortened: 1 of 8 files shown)


array.go
```diff
@@ -352,32 +347,51 @@ func (a *Array) NonEmptyDomain() ([]NonEmptyDomain, bool, error) {
...
 		}
+
+		nonEmptyDomains := make([]NonEmptyDomain, 0)
+		for i := uint(0); i < ndims; i++ {
+			dimension, err := domain.DimensionFromIndex(i)
+			if err != nil {
+				return nil, false, err
+			}
+
+			nonEmptyDomain, err := getNonEmptyDomainForDim(dimension, tmpDomain)
```
Shelnutt2, line 362: I believe you'll have to iterate through `tmpDomain` here, else you'll end up with the same values. It might be easier to drop `tiledb_array_get_non_empty_domain` and switch to `tiledb_array_get_non_empty_domain_from_index` inside the for loop?

serialize.go
```diff
@@ -138,13 +138,23 @@ func DeserializeArrayNonEmptyDomain(a *Array, buffer *Buffer, serializationType
...
+	}
+
+	nonEmptyDomains := make([]NonEmptyDomain, 0)
+	for i := uint(0); i < ndims; i++ {
+		dimension, err := domain.DimensionFromIndex(i)
+		if err != nil {
+			return nil, false, err
+		}
+
+		nonEmptyDomain, err := getNonEmptyDomainForDim(dimension, tmpDomain)
```
Shelnutt2, line 150: Same as above you will need to iterate through `tmpDomain` here, else you'll end up with the same values.

array.go
antalakas: `tmpDomain` is just a slice here. Initially it was used in a call to `makeNonEmptyDomain` which was receiving a domain and did a loop in dimensions. Here we just use `getNonEmptyDomainForDim` which accepts a diimension and we are performing the loop through domain dimensions around this function. I don't think i changed the intention of the initial code. I will add a test for `tiledb_array_get_non_empty_domain` to make sure anyway

Shelnutt2: The tmpDomain is created with `tmpDomain, tmpDomainPtr, err := domainType.MakeSlice(uint64(2 * ndims))`. So it is the size of `2*ndims`. In the new `getNonEmptyDomainForDim` you access index 0 and 1 of the passed slice. The problem I see is you always pass the same slice, `tmpDomain` so for every dimension you are just accessing `tmpDomain[0]` and `tmpDomain[1]`. You need to subslice tmpDomain based on the index dimension, however that would require you to add switch cases for all the datatypes. I think it might be easier to switch to `tiledb_array_get_non_empty_domain_by_index` to just fetch the single dimension's nonEmptyDomain to pass to `getNonEmptyDomainForDim`

antalakas: Got it. For this case, using the indexed version works well. In the case of `tiledb_deserialize_array_nonempty_domain` there is no indexed version so there is no choice but using the switch

Shelnutt2:
Thanks for the changes, looks great!

PR #109, opened 2020-04-12
adds dimensions support in set buffer

antalakas:


query.go
```diff
@@ -208,8 +208,11 @@ func (q *Query) SetBufferUnsafe(attribute string, buffer unsafe.Pointer, bufferS
 
 // SetBuffer Sets the buffer for a fixed-sized attribute to a query
 // The buffer must be an initialized slice
-func (q *Query) SetBuffer(attribute string, buffer interface{}) (*uint64,
+func (q *Query) SetBuffer(attributeOrDimension string, buffer interface{}) (*uint64,
 	error) {
+
+	fmt.Printf("%v\n", attributeOrDimension)
```
Shelnutt2, line 214: I believe this line is what is causing the unit tests to fail.

query.go
```diff
@@ -950,6 +986,7 @@ func (q *Query) SetBufferVar(attribute string, offset []uint64, buffer interface
 		// Store slice so underlying array is not gc'ed
 		q.buffers = append(q.buffers, tmpBuffer)
 		cbuffer = unsafe.Pointer(&(tmpBuffer)[0])
+		// cbuffer = unsafe.Pointer(C.CString("ccbbddddaa"))
```
Shelnutt2, line 989: remove?

PR #110, opened 2020-04-13
Support variable sized ranges

antalakas:


query.go
```diff
@@ -518,6 +518,40 @@ func (q *Query) AddRange(dimIdx uint32, start interface{}, end interface{}) erro
 	return nil
 }
 
+// AddRangeVar adds a range applicable to variable-sized dimensions
+// Applicable only to string dimensions
+func (q *Query) AddRangeVar(dimIdx uint32, start string, end string) error {
```
Shelnutt2, line 523: Have this take an interface not a string so we are future proofed from the API perspective. We can then put the `tiledb_query_add_range_var` range function inside a switch case on the dimension type. We can just leave only strings supported and error for anything else.

I believe its better to support it this way so in the future we don't break the API of taking interface instead of strings.

antalakas: Agreed!

antalakas: Done

PR #111, opened 2020-04-13
Support non empty domain var

antalakas:


array.go
```diff
@@ -360,6 +360,97 @@ func (a *Array) NonEmptyDomain() ([]NonEmptyDomain, bool, error) {
...
+	var isEmpty C.int32_t
+
+	var start interface{}
+	var end interface{}
+	var cstart unsafe.Pointer
+	var cend unsafe.Pointer
+
+	switch domainType {
+	case TILEDB_STRING_ASCII:
+		ret := C.tiledb_array_get_non_empty_domain_var_size_from_name(
```
Shelnutt2, line 398: I think everything can be moved outside the switch case here? There is nothing specific to strings?

array.go
```diff
@@ -360,6 +360,97 @@ func (a *Array) NonEmptyDomain() ([]NonEmptyDomain, bool, error) {
...
+	if err != nil {
+		return nil, nil, false, err
+	}
+
+	domain, err := schema.Domain()
+	if err != nil {
+		return nil, nil, false, err
+	}
+
+	domainType, err := domain.Type()
```
Shelnutt2, line 378: Domain type can not be used here. We should no longer use domain type anywhere anywhere. Its is only valid for homogeneous dimensions. We shouldn't even try to use it though because using the specific dimension datatype will work in all cases.

array.go
```diff
@@ -360,6 +360,97 @@ func (a *Array) NonEmptyDomain() ([]NonEmptyDomain, bool, error) {
 	}
 }
 
+// NonEmptyDomainVarFromName retrieves the non-empty domain from an array for a
+// given dimension name.
+// This returns the bounding coordinates for the dimension
+func (a *Array) NonEmptyDomainVarFromName(dimName string) ([]NonEmptyDomain, []NonEmptyDomain, bool, error) {
```
Shelnutt2, line 366: We should not need to return two lists of nonEmptyDomains here. This should just return a single `NonEmptyDomain` object no?

array.go
```diff
@@ -360,6 +360,97 @@ func (a *Array) NonEmptyDomain() ([]NonEmptyDomain, bool, error) {
...
+
+		if isEmpty == 1 {
+			return nil, nil, true, nil
+		} else {
+			nonEmptyDomainsStart, err := makeNonEmptyDomain(domain, start)
+			if err != nil {
+				return nil, nil, false, err
+			}
+
+			nonEmptyDomainsEnd, err := makeNonEmptyDomain(domain, end)
```
Shelnutt2, line 442: We have to include the changes for `makeNonEmptyDomain` in this branch to support the var domains.

array.go
antalakas: Got it

array.go
antalakas: Correct it is improved in https://github.com/TileDB-Inc/TileDB-Go/pull/108 just merged in this branch

Shelnutt2:
@antalakas can you also add support for `tiledb_array_get_non_empty_domain_var_from_index`?

array.go
antalakas: included

array.go
antalakas: moved in merged branch

antalakas:
> `tiledb_array_get_non_empty_domain_var_from_index`

@Shelnutt2 You mean as a standalone function?

Shelnutt2:
> > `tiledb_array_get_non_empty_domain_var_from_index`
> 
> @Shelnutt2 You mean as a standalone function?

Yes, in this PR you implemented the `ByName` function, we need to also implement the `ByIndex` function

antalakas:
> > > `tiledb_array_get_non_empty_domain_var_from_index`
> > 
> > 
> > @Shelnutt2 You mean as a standalone function?
> 
> Yes, in this PR you implemented the `ByName` function, we need to also implement the `ByIndex` function

Done

PR #112, opened 2020-04-15
Support hetereogeneus dimensions

antalakas:
Should we drop `TILEDB_COORDS`?

### DIFF
--- a/examples/writing_sparse_heter_dim_test.go
+++ b/examples/writing_sparse_heter_dim_test.go
@@ -121,6 +121,15 @@ func readSparseHeterDimArray() {
 	err = array.Open(tiledb.TILEDB_READ)
 	checkError(err)
 
+	// Non-empty domain: [1,4], [1,4]
+	x, isEmpty, err := array.NonEmptyDomain()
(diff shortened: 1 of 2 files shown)


examples/writing_sparse_heter_dim_test.go
```diff
@@ -0,0 +1,172 @@
...
+	checkError(err)
+	err = array.Open(tiledb.TILEDB_READ)
+	checkError(err)
+
+	buffD1R := make([]float32, 4)
+	buffD2R := make([]int64, 4)
+	buffAR := make([]int32, 4)
+
+	// Prepare the query
+	query, err := tiledb.NewQuery(ctx, array)
```
Shelnutt2, line 129: For completeness can you add a a check for the nonEmptyDomain too?

antalakas: Sure, done

Shelnutt2:
> Should we drop `TILEDB_COORDS`?

Not yet, it will be deprecated in core with 2.0 but let's wait until its removed in 2.1 to remove support here.

PR #116, opened 2020-04-28
Uses tiledb_domain_has_dimension

antalakas:


query.go
```diff
@@ -255,6 +255,15 @@ func (q *Query) SetBuffer(attributeOrDimension string, buffer interface{}) (*uin
 					err)
 			}
 
+			hasDim, err := domain.HasDimension(attributeOrDimension)
```
Shelnutt2, line 258: This should be above, the idea is to not call `schema.AttributeFromName` if its a dimension. We should check if its a dim or a attr before trying to fetch the object. This apply everywhere.

antalakas: According to TileDB export comments: `tiledb_domain_has_dimension` : `Checks whether the domain has a dimension of the given name.` Will it work with attribute names?

ihnorton: There is also `tiledb_array_schema_has_attribute`.

query.go
```diff
@@ -227,27 +227,31 @@ func (q *Query) SetBuffer(attributeOrDimension string, buffer interface{}) (*uin
...
 		}
 	} else {
-		schemaAttribute, err := schema.AttributeFromName(attributeOrDimension)
+		hasDim, err := domain.HasDimension(attributeOrDimension)
 		if err != nil {
+			return nil, err
+		}
+
+		if hasDim {
 			domain, err := schema.Domain()
```
Shelnutt2, line 255: you can remove domain here, you fetched it above

PR #119, opened 2020-05-13
Add timestamp support to getMetadataValue

antalakas:


array.go
```diff
@@ -1272,6 +1273,57 @@ func (a *Array) ConsolidateMetadataWithKey(encryptionType EncryptionType, key st
...
 
+// GetTimeFromTimestamp returns a time.Time object for a time related TileDB datatype
+func GetTimeFromTimestamp(datatype Datatype, timestamp uint64) time.Time {
+	var then time.Time
+	switch datatype {
+	case TILEDB_DATETIME_YEAR:
+		then = time.Date(int(timestamp), 1, 1, 0, 0, 0, 0, time.UTC)
+	case TILEDB_DATETIME_MONTH:
+		now := time.Now()
+		then = time.Date(now.Year(), time.Month(timestamp), 1, 0, 0, 0, 0, time.UTC)
```
Shelnutt2, line 1284: The datetimes in tiledb are deltas from unix epoch with a resolution of the specified time. So month isn't just 1-12. You need to convert the number of months since epoch into the datetime. Same of all datetime types.

array.go
```diff
@@ -1394,6 +1394,17 @@ func getMetadataValue(datatype Datatype, valueNum uint, cvalue unsafe.Pointer) (
...
 		value = C.GoString(&tmpslice[0])[0:valueNum]
+	case TILEDB_DATETIME_YEAR, TILEDB_DATETIME_MONTH, TILEDB_DATETIME_WEEK,
+		TILEDB_DATETIME_DAY, TILEDB_DATETIME_HR, TILEDB_DATETIME_MIN,
+		TILEDB_DATETIME_SEC, TILEDB_DATETIME_MS, TILEDB_DATETIME_US,
+		TILEDB_DATETIME_NS, TILEDB_DATETIME_PS, TILEDB_DATETIME_FS,
+		TILEDB_DATETIME_AS:
+		if valueNum > 1 {
+			return nil, fmt.Errorf("Unrecognized value type: %d", datatype)
+		} else {
+			var timestamp interface{} = *(*uint16)(unsafe.Pointer(cvalue))
```
Shelnutt2, line 1405: should this be a uint64 ?

antalakas: Fixed

array.go
antalakas: Changed the logic

epoch.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	} else {
+		leapFlag = false
+	}
+	return leapFlag
+}
+
+func secondsFromEpochYears(numOfYears int64) int64 {
+	var numOfSeconds int64 = 0
+	for y := int64(epochYear); y < epochYear+numOfYears; y++ {
```
Shelnutt2, line 32: I forgot to mention we need to handle negative values too, this is how we get pre-epoch time.

antalakas: Ok

PR #121, opened 2020-07-13
Handles TILEDB_CHAR array metadata

antalakas:


array.go
```diff
@@ -1386,7 +1404,7 @@ func getMetadataValue(datatype Datatype, valueNum uint, cvalue unsafe.Pointer) (
 			value = *(*float64)(unsafe.Pointer(cvalue))
 		}
 	case TILEDB_CHAR:
-		tmpslice := (*[1 << 46]C.char)(unsafe.Pointer(cvalue))[:valueNum:valueNum]
+		tmpslice := (*[1 << 30]C.char)(unsafe.Pointer(cvalue))[:valueNum:valueNum]
```
Shelnutt2, line 1407: Is there a reason we can't use 46 for the size of the array here and are limited to 32 bit size?

antalakas: It does using 46 works too here and it seems that i still cannot reproduce the error

antalakas: It seems that the only way to have a segfault here is for `cvalue` to be `nil`

antalakas: Checked with vcf-sample-20/data array and reproduced the error. Indeed for index 4 `cvalue` is `null` pointer, `valueNum=1`. Can this be a bug in core?

Shelnutt2: What is the metadata key for the value that is null?

antalakas: Screenshot 2020-07-14 at 12 35 57 AM

PR #122, opened 2020-08-02
Add support for io.Writer/Reader/Seeker in VFSfh

Shelnutt2:
This allows a VFS file handle to be used in functions which take one of the io package interfaces

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -11,6 +11,7 @@ import "C"
 
 import (
 	"fmt"
+	"io"
 	"runtime"
 	"strings"
 	"unsafe"
@@ -22,8 +23,12 @@ const arrayMetadataFolderName = "__meta"
 
 // VFSfh is a virtual file system file handler
 type VFSfh struct {
+	vfs         *VFS
 	tiledbVFSfh *C.tiledb_vfs_fh_t
 	context     *Context
+	offset      uint64
+	size        *uint64
+	uri         string
 }
 
 // Free a tiledb c vfs file handler
@@ -320,7 +325,7 @@ func (v *VFS) MoveDir(oldURI string, newURI string) error {
 func (v *VFS) Open(uri string, mode VFSMode) (*VFSfh, error) {
 	curi := C.CString(uri)
 	defer C.free(unsafe.Pointer(curi))
-	fh := &VFSfh{context: v.context}
+	fh := &VFSfh{context: v.context, uri: uri, vfs: v}
 	// Set finalizer for free C pointer on gc
 	runtime.SetFinalizer(fh, func(fh *VFSfh) {
 		fh.Free()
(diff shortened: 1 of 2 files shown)


vfs.go
```diff
@@ -396,3 +399,107 @@ func (v *VFS) DirSize(uri string) (uint64, error) {
...
+		}
+		v.size = &size
+	}
+
+	// If the requested read size is beyond the limit, adjust bytes to read
+	if v.offset+nbytes > *v.size {
+		nbytes = *v.size - v.offset
+	}
+
+	if nbytes == 0 {
```
antalakas, line 436: Why not check this first and exit early?

vfs.go
```diff
@@ -396,3 +399,107 @@ func (v *VFS) DirSize(uri string) (uint64, error) {
...
+
+	v.offset += nbytes
+	return int(nbytes), nil
+}
+
+// Write the contents of a buffer into a file. Note that this function only
+// appends data at the end of the file. If the file does not exist,
+// it will be created
+func (v *VFSfh) Write(bytes []byte) (int, error) {
+	cbuffer := C.CBytes(bytes)
```
antalakas, line 455: `C.CBytes ` or `unsafe.Pointer(&bytes[0])`

vfs.go
```diff
@@ -396,3 +399,107 @@ func (v *VFS) DirSize(uri string) (uint64, error) {
...
+
+	switch whence {
+	case io.SeekStart:
+		origin = 0
+		break
+	case io.SeekCurrent:
+		origin = v.offset
+		break
+	case io.SeekEnd:
+		origin = *v.size
```
antalakas, line 492: Are we sure that `v.size` is set?

antalakas: Also origin is 0 after `var origin uint64` so `SeekEnd` will be 0 always right?

antalakas (changes requested):

vfs.go
Shelnutt2: We can't check this first incase we need to set the nbytes from the size - offset.

vfs.go
Shelnutt2: Thanks, adjusted

snagles (approved):

vfs.go
```diff
@@ -462,3 +465,112 @@ func (v *VFS) NumOfFragmentsInPath(path string) (int, error) {
...
+	return nil
+}
+
+// Read part of a file
+func (v *VFSfh) Read(p []byte) (int, error) {
+	nbytes := uint64(len(p))
+
+	// If the size is empty, fetch it
+	if v.size == nil {
+		size, err := v.vfs.FileSize(v.uri)
```
antalakas, line 490: This happens in two places, maybe a wrapper function is a good choice

vfs.go
```diff
@@ -462,3 +465,112 @@ func (v *VFS) NumOfFragmentsInPath(path string) (int, error) {
...
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Unknown error in VFS.Close: %s", v.context.LastError())
+	}
+
+	v.Free()
+	return nil
+}
+
+// Read part of a file
+func (v *VFSfh) Read(p []byte) (int, error) {
```
antalakas, line 485: Read and Write signatures end up the same and this confuses me. I don't know if the out type argument is used in go, but i 'd prefer to allocate memory in this function taking the number of bytes needed and returning the []byte

antalakas (approved):

vfs.go
Shelnutt2: We can't edit the Read/Write signatures they are part of the interface(s): https://golang.org/pkg/io/#Reader

vfs.go
Shelnutt2: Good idea

PR #125, opened 2020-10-13
Support vacuum

antalakas:
[ch3306]

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -2,17 +2,21 @@ package tiledb
 
 /*
 #cgo LDFLAGS: -ltiledb
-#cgo linux LDFLAGS: -ldl
+#cgo linux LDFLAGS: -ldl -lclibrary
 #include <tiledb/tiledb.h>
 #include <stdlib.h>
+#include "clibrary.h"
 */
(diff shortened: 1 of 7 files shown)


vfs.go
```diff
@@ -396,3 +400,58 @@ func (v *VFS) DirSize(uri string) (uint64, error) {
...
+// Callback is a type
+type Callback struct {
+	Func func(string, unsafe.Pointer)
+	Path string
+	Data unsafe.Pointer
+}
+
+//export numOfFragmentsInPath
+func numOfFragmentsInPath(path *C.cchar_t, data unsafe.Pointer) int32 {
+	ctx, err := NewContext(nil)
```
antalakas, line 413: vfs needs to come from data pointer

Shelnutt2: You should also pass in the config somehow because if the user sets credentials or configs those won't be set here.

antalakas: Provided that vfs comes from the caller, they will have the opportunity to set config when creating vfs object right?

Shelnutt2 (approved):

PR #133, opened 2020-11-19
Add timestamp correlation test

antalakas:
When run, this test will create a 2D dense array, write data and metadata
to it multiple times keeping track of write timestamps, then read at specific
timestamp using array.OpenAt and prove correlation of timestamps to fragments
and metadata created /updated

### DIFF
--- a/examples/reading_timestamp_test.go
+++ b/examples/reading_timestamp_test.go
@@ -1,0 +1,248 @@
+/**
+ * @file   reading_timestamp_test.go
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2020 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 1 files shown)


examples/reading_timestamp_test.go
```diff
@@ -0,0 +1,247 @@
...
+	checkError(err)
+
+	// Prepare some data for the array
+	data := []int32{
+		1, 2 + bias, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16}
+
+	// Open the array for writing and create the query.
+	array, err := tiledb.NewArray(ctx, timestampArrayName)
+	checkError(err)
+	err = array.Open(tiledb.TILEDB_WRITE)
```
Shelnutt2, line 101: You should use OpenAt for the writes too if you want to force a given timestamp and make the test for deterministic.

antalakas: I guess the value passed to timestamp would be `time.Now`. Isn't it the same as just Open as currently written?

Also based on test output behavior seems deterministic even though in the case we 

1. write data + metadata
2. write metadata
3. write metadata

only one fragment is created

I think i am missing something here

antalakas: @Shelnutt2 updated

Shelnutt2 (approved):

PR #140, opened 2020-11-26
Update README with proper installation link

davedbase:
URLs are broken tried updating the ones I could find, however Quickstart link in Quick Links sections doesn't have an equivalent.


Shelnutt2 (approved):
Thanks for the PR and updating the links!

PR #144, opened 2020-12-11
Fix object management

antalakas:
Object is not a struct anymore, we rather handle object management by means of global functions in tiledb package

### DIFF
--- a/object.go
+++ b/object.go
@@ -18,14 +18,19 @@ import (
 // ObjectType returns the object type
 // A TileDB "object" is currently either a TileDB array or a TileDB group.
 func ObjectType(ctx *Context, path string) (ObjectTypeEnum, error) {
+	if ctx == nil {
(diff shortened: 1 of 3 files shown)


object.go
```diff
@@ -9,47 +9,29 @@ package tiledb
...
-	}
-	return &object, nil
-}
-
-// Type returns the object type
-func (o *Object) Type() (ObjectType, error) {
-	var objectType C.tiledb_object_t
-	cpath := C.CString(o.path)
+func ObjectType(ctx *Context, path string) (ObjectTypeEnum, error) {
+	var objectTypeEnum C.tiledb_object_t
```
Shelnutt2, line 21: Lets add a context != nil check to all the functions

Shelnutt2 (approved):
One comment, besides that lgtm

object.go
antalakas: 👍

PR #149, opened 2021-01-21
Add two nullable attribute functions for buffer size

snagles:


query_test.go
```diff
@@ -594,6 +597,8 @@ func TestQueryEffectiveBufferSizeHeterogeneous(t *testing.T) {
 	assert.Nil(t, err)
 	_, err = query.SetBuffer("cols", colsWrite)
 	assert.Nil(t, err)
+	_, _, _, err = query.SetBufferVarNullable("a2", a2OffWrite, a2DataWrite, a2Validity)
```
Shelnutt2, line 600: Add a set for `SetBufferNullable`/`BufferSizeVarNullable` too?

query.go
```diff
@@ -2090,6 +2090,69 @@ func (q *Query) BufferSizeVar(attributeOrDimension string) (uint64, uint64, erro
 	return offsetNumElements, dataNumElements, nil
 }
 
+// BufferSizeVarNullable returns the size (in num elements) of the backing C buffers for the given variable-length nullable attribute
+func (q *Query) BufferSizeVarNullable(attributeOrDimension string) (uint64, uint64, uint64, error) {
```
antalakas, line 2094: Since nullable support is only for attributes maybe you need to set attributeName or so as an argument

query.go
```diff
@@ -2162,6 +2225,59 @@ func (q *Query) BufferSize(attributeNameOrDimension string) (uint64, error) {
 	return dataNumElements, nil
 }
 
+// BufferSizeNullable returns the size (in num elements) of the backing C buffer for the given nullable attribute
+func (q *Query) BufferSizeNullable(attributeNameOrDimension string) (uint64, uint64, error) {
```
antalakas, line 2229: Since nullable support is only for attributes maybe you need to set attributeName or so as an argument

query_test.go
snagles: Good idea, done

antalakas (approved):

query.go
Shelnutt2: The below code is also wrong because of this. You don't fetch the datatype of the attribute but actually looks like it unset except for `TILEDB_COORS`? That is an impossible condition cause `TILEDB_COORDS` can't be nullable.

snagles: Good catch, fixed

query.go
snagles: Good catch, fixed

query.go
```diff
@@ -2090,6 +2090,69 @@ func (q *Query) BufferSizeVar(attributeOrDimension string) (uint64, uint64, erro
...
 
+// BufferSizeVarNullable returns the size (in num elements) of the backing C buffers for the given variable-length nullable attribute
+func (q *Query) BufferSizeVarNullable(attributeName string) (uint64, uint64, uint64, error) {
+	var datatype Datatype
+	schema, err := q.array.Schema()
+	if err != nil {
+		return 0, 0, 0, err
+	}
+
+	domain, err := schema.Domain()
```
Shelnutt2, line 2101: The below code is also wrong because of this. You don't fetch the datatype of the attribute but actually looks like it unset except for TILEDB_COORS? That is an impossible condition cause TILEDB_COORDS can't be nullable. This should be handling and checking attributes.

query.go
```diff
@@ -2162,6 +2225,59 @@ func (q *Query) BufferSize(attributeNameOrDimension string) (uint64, error) {
...
 
+// BufferSizeNullable returns the size (in num elements) of the backing C buffer for the given nullable attribute
+func (q *Query) BufferSizeNullable(attributeName string) (uint64, uint64, error) {
+	var datatype Datatype
+	schema, err := q.array.Schema()
+	if err != nil {
+		return 0, 0, err
+	}
+
+	domain, err := schema.Domain()
```
Shelnutt2, line 2236: This function is also wrong because of attribute not dimensions.

Shelnutt2 (changes requested):

antalakas (approved):

query_test.go
```diff
@@ -576,6 +576,11 @@ func TestQueryEffectiveBufferSizeHeterogeneous(t *testing.T) {
 	colsWrite := []int64{1, 1, 2}
 	a1DataWrite := []byte("abbccc")
 	a1OffWrite := []uint64{0, 1, 3}
+	a2DataWrite := []byte("abbcccdddd")
```
Shelnutt2, line 579: `a2` and `a3` are missing from the schema.

Shelnutt2 (changes requested):

query_test.go
snagles: Added

query_test.go
```diff
@@ -550,6 +558,12 @@ func TestQueryEffectiveBufferSizeHeterogeneous(t *testing.T) {
 	err = arraySchema.AddAttributes(attribute)
 	assert.Nil(t, err)
 
+	err = arraySchema.AddAttributes(attribute2)
+	assert.Nil(t, err)
+
+	err = arraySchema.AddAttributes(attribute3)
```
antalakas, line 564: Does it need to be set as nullable?

antalakas (changes requested):

query_test.go
Shelnutt2: Yes both need to be set to nullable, and a2 need to be set to var length also.

snagles: Done

Shelnutt2 (approved):

antalakas (changes requested):

antalakas (approved):

PR #150, opened 2021-01-22
Use Github Actions

antalakas:
Removes Azure Pipelines in favor of Github Actions, fixes tests that were previously not failing in CI


query_test.go
```diff
@@ -1151,20 +1151,20 @@ func TestQueryReadEmpty(t *testing.T) {
 	// Submit query
 	assert.Nil(t, query.Submit())
 
-	// Validate status, since query was used this is should be complete
+	// Validate status, since array was empty this is should be incomplete
 	status, err := query.Status()
 	assert.Nil(t, err)
-	assert.Equal(t, TILEDB_COMPLETED, status)
+	assert.Equal(t, TILEDB_INCOMPLETE, status)
```
antalakas, line 1157: @Shelnutt2 this needs to be incomplete for test to pass, has the behavior changed in the case of reading an empty array?

query_test.go
```diff
@@ -1151,20 +1151,20 @@ func TestQueryReadEmpty(t *testing.T) {
...
 	queryType, err := query.Type()
 	assert.Nil(t, err)
 	assert.Equal(t, TILEDB_READ, queryType)
 
-	// No results because query it empty
+	// Has results because array it empty
 	hasResults, err := query.HasResults()
 	assert.Nil(t, err)
-	assert.Equal(t, false, hasResults)
+	assert.Equal(t, true, hasResults)
```
antalakas, line 1167: @Shelnutt2 has the behavior changed in the case of reading an empty array? I can see results in offsets after query submit, stepping at the var sized attribute data type size, actual values zero

query_test.go
Shelnutt2: No, behavior change in core. Incomplete is an error here. I think it might be the size of the buffers, try increasing them slightly and see. With such small buffer sometimes the estimation in core about if it can run a query or not is off from what is expected.

query_test.go
Shelnutt2: No behavior change

snagles (approved):
LGTM, that was quick, nice job!

Shelnutt2 (approved):

PR #159, opened 2021-03-30
Set finalizers for misc objs

snagles:


Shelnutt2 (approved):

antalakas (approved):

context.go
```diff
@@ -71,6 +71,11 @@ func (c *Context) Config() (*Config, error) {
 		return nil, fmt.Errorf("Unknown error in GetConfig")
 	}
 
+	// Set finalizer for free C pointer on gc
```
antalakas, line 74: It seems get config from `NewConfig` and rely on `SetFinalizer` in there
	
```
config, err := NewConfig()
if err != nil {
   return nil, err
}
```

passes CI

snagles: Good catch, done

PR #160, opened 2021-03-30
Call free on all char allocs

snagles:


antalakas (approved):

config_iter.go
```diff
@@ -56,6 +56,9 @@ func (ci *ConfigIter) Free() {
 func (ci *ConfigIter) Here() (*string, *string, error) {
 	var err *C.tiledb_error_t
 	var cparam, cvalue *C.char
+	defer C.free(unsafe.Pointer(cparam))
```
Shelnutt2, line 59: This is wrong, we don't own these strings its owned by the config iterator in C. We must not free them.

context.go
```diff
@@ -92,9 +91,9 @@ func (c *Context) LastError() error {
 
 	if err != nil {
 		var msg *C.char
+		defer C.free(unsafe.Pointer(msg))
```
Shelnutt2, line 94: This is wrong, we don't own these strings its owned by the error message in C, we must not free it.

query.go
```diff
@@ -3205,6 +3205,7 @@ func (q *Query) GetFragmentNum() (*uint32, error) {
 // GetFragmentURI returns uri for a fragment
 func (q *Query) GetFragmentURI(num uint64) (*string, error) {
 	var cURI *C.char
+	defer C.free(unsafe.Pointer(cURI))
```
Shelnutt2, line 3208: We don't own the cURI here, it can't be freeded.

Shelnutt2 (changes requested):
I don't think anything here can be merged.

Shelnutt2:
@snagles please revert this is wrong.

PR #167, opened 2021-04-20
Add config compare and query config getter

Shelnutt2:
Add config compare and query config getter


antalakas (approved):
Is new version of TileDB required for CI?

PR #171, opened 2021-05-19
automatically generate a file with byte counts to reduce unsafe.Sizeof verbosity.

thetorpedodog:
These two changes:

1. create a go:generate tool to make a file with the byte sizes of all the numeric datatypes, and
2. run the tool and replace all other uses of unsafe.Sizeof with those constants.

The line count is large but it’s less scary than it looks—much of it is testdata and find-and-replace.

Right now I have it set up to export all the constants (and the map) that it creates, but I am open to making it package-private to shrink the API surface. However, internal code seems to recalculate `unsafe.Sizeof` stuff a lot, so it seems like it would be useful to external users too.


thetorpedodog:
Another idea I had after getting this initial version done was to put the generated file into its own directory, maybe `bytesizes`. That would keep from adding another bunch of entries to the `tiledb` namespace (especially since this is a basic thing with no dependencies on anything else) and allow shortening the names (`bytesizes.Uint8`) while still making the usage sites highly readable. A commit for that is forthcoming.

snagles:
I like this a lot and I agree with the implementation as well. I think there are some other areas for de-duplication/simplification in a similar vein as this, nice job. 

I'm good with merging this assuming no opposition by @antalakas  or @Shelnutt2

PR #172, opened 2021-05-21
Use consistent naming for a tiledb context type

snagles:
Uses the same naming schema for the tiledb context type


thetorpedodog:
I am strongly in favor of this because it prevents confusion with the built-in `context.Context` type. Code looks good.

thetorpedodog (approved):
submit imo; this is trivial and makes no change to the public api and I think even the literal bytes of the binary will be unchanged

PR #173, opened 2021-05-25
Update TileDB to 2.3

snagles:
- Adds support for opening with timestamp start/end options using a new OpenWithOptions function
- Adds Getters for retrieval that is retrieved directly from core as the state is managed by core itself and not as an array property
- Updates tests for slight changes in inputs/outputs
- Adds new tiledb time types
- Add query condition logic


query_test.go
```diff
@@ -1952,7 +1946,7 @@ func TestSparseQueryWriteNullable(t *testing.T) {
 
 	// Create write buffers
 	bufferA1 := []int32{1, 2, 3}
-	bufferA1Validity := []uint8{1, 1, 0}
+	bufferA1Validity := []uint8{1, 1, 0, 0}
```
Shelnutt2, line 1949: Why is there 4 validity values here but the data buffer only has 3 values?

snagles: Looks like the read buffer had an extra value leading to the mismatch, I've updated bufferA1 to have a fourth value to reflect readBufferA1Validity

Shelnutt2: I don't think 4 values is correct, there is only 3 values set for dim1. Why is this not erroring out with a mismatch between the number of cells being written?

snagles: That I'm not sure. It passes if I set them all back to 3 values

snagles: It seems to me like this was not errored in the past and the test was wrong with the value of 4. New TileDB caught this an did error. So it was a mistake in the test (and in me changing the others to 4)

Shelnutt2 (approved):

Shelnutt2 (approved):

query_condition.go
```diff
@@ -0,0 +1,198 @@
...
+
+	switch v := value.(type) {
+	case string:
+		valueSize := C.uint64_t(len(v))
+		cTmpValue := C.CString(v)
+		defer C.free(unsafe.Pointer(cTmpValue))
+		if ret := C.tiledb_query_condition_init(qc.context.tiledbContext, qc.cond, cname, unsafe.Pointer(cTmpValue), valueSize, C.tiledb_query_condition_op_t(op)); ret != C.TILEDB_OK {
+			return fmt.Errorf("Error initing tiledb string query condition: %s", qc.context.LastError())
+		}
+	case int:
```
antalakas, line 85: So, two cases are needed right? One for scalars, one for slices

query_condition.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	case float32:
+		if ret := C.tiledb_query_condition_init(qc.context.tiledbContext, qc.cond, cname, unsafe.Pointer(&v), C.uint64_t(size), C.tiledb_query_condition_op_t(op)); ret != C.TILEDB_OK {
+			return fmt.Errorf("Error initing tiledb numeric query condition: %s", qc.context.LastError())
+		}
+	case float64:
+		if ret := C.tiledb_query_condition_init(qc.context.tiledbContext, qc.cond, cname, unsafe.Pointer(&v), C.uint64_t(size), C.tiledb_query_condition_op_t(op)); ret != C.TILEDB_OK {
+			return fmt.Errorf("Error initing tiledb numeric query condition: %s", qc.context.LastError())
+		}
+	case []int:
```
antalakas, line 129: Case for slices

vfs.go
```diff
@@ -485,7 +485,7 @@ func (v *VFSfh) Close() error {
 }
 
 // Read part of a file
-func (v *VFSfh) Read(p []byte) (uint64, error) {
+func (v *VFSfh) Read(p []byte) (int, error) {
```
antalakas, line 488: Why are we changing the return type here since all usage in function is of uint64?

query_condition_test.go
```diff
@@ -0,0 +1,414 @@
...
+			assert.Nil(t, err)
+			assert.NotNil(t, query)
+
+			qc1, err := NewQueryCondition(array.context, "a1", c.op1, c.op1Value)
+			assert.Nil(t, err)
+
+			qc2, err := NewQueryCondition(array.context, "a1", c.op2, c.op2Value)
+			assert.Nil(t, err)
+
+			qc, err := NewQueryConditionCombination(array.context, qc1, c.combinationCondition, qc2)
```
antalakas, line 132: Since `NewQueryConditionCombination` does not alter existing conditions, isn't this line's effect shadowed by above two lines (`NewQueryCondition`)?

antalakas (approved):
LGTM!

vfs.go
snagles: This was because of the signature of the reader interface: https://golang.org/pkg/io/#Reader

It was a mistake for me to have changed this previously

query_condition_test.go
snagles: The NewQueryConditionCombination functionality just merges the two conditions together and doesn't mutate the underlying query conditions. In effect it creates a new query condition with both predicates in one "query condition". We then use only this query condition when doing the actual query

antalakas: 👍

vfs.go
antalakas: 👍

PR #174, opened 2021-05-26
Use sized values where applicable

snagles:


thetorpedodog:
This seems fine but I don't understand the motivation.

thetorpedodog:
It looks like you have files in `examples/...` that should not exist.

thetorpedodog (approved):
looks a-good

snagles:
> It looks like you have files in `examples/...` that should not exist.

good catch, ty

PR #179, opened 2021-06-24
Return empty json object bytes if empty

snagles:
Return an empty json object in bytes if tiledb does not error

### DIFF
--- a/context.go
+++ b/context.go
@@ -151,17 +151,21 @@ func (c *Context) setDefaultTags() error {
 	return nil
 }
 
-// Stats gets stats for a context as a string
-func (c *Context) Stats() (string, error) {
+// Stats gets stats for a context as json bytes
+func (c *Context) Stats() ([]byte, error) {
 	var stats *C.char
 	if ret := C.tiledb_ctx_get_stats(c.tiledbContext, &stats); ret != C.TILEDB_OK {
-		return "", fmt.Errorf("Error getting stats from context: %s", c.LastError())
+		return nil, fmt.Errorf("Error getting stats from context: %s", c.LastError())
 	}
 
 	s := C.GoString(stats)
 	if ret := C.tiledb_stats_free_str(&stats); ret != C.TILEDB_OK {
-		return "", fmt.Errorf("Error freeing string from dumping stats to string")
+		return nil, fmt.Errorf("Error freeing string from dumping stats to string")
 	}
 
-	return s, nil
+	if s == "" {
+		return []byte("{}"), nil
+	}
+
+	return []byte(s), nil
 }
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -151,17 +151,21 @@ func (c *Context) setDefaultTags() error {
 	return nil
 }
 
-// Stats gets stats for a context as a string
-func (c *Context) Stats() (string, error) {
+// Stats gets stats for a context as json
```
thetorpedodog, line 154: 1. I would describe this as "JSON bytes" to be abundantly clear.
2. Why the switch from a string to bytes?

context.go
```diff
@@ -151,17 +151,21 @@ func (c *Context) setDefaultTags() error {
...
 
 	s := C.GoString(stats)
 	if ret := C.tiledb_stats_free_str(&stats); ret != C.TILEDB_OK {
-		return "", fmt.Errorf("Error freeing string from dumping stats to string")
+		return nil, fmt.Errorf("Error freeing string from dumping stats to string")
 	}
 
-	return s, nil
+	if s == "" {
+		return []byte("{}"), nil
```
thetorpedodog, line 167: should this be `{}` or `null`? (either might be fine, just want to make sure we're doing the right thing here)

snagles: I think either is okay, I think null didn't use to be valid according to json spec but it looks like this has changed

context.go
snagles: 1. Updated comment
2. Switching to bytes provides more compatibility with the normal encoding/json package as it is the common container for most actions, e.g things like json.RawMessage, decoding etc

context.go
```diff
@@ -151,17 +151,21 @@ func (c *Context) setDefaultTags() error {
...
 }
 
-// Stats gets stats for a context as a string
-func (c *Context) Stats() (string, error) {
+// Stats gets stats for a context as json bytes
+func (c *Context) Stats() ([]byte, error) {
 	var stats *C.char
 	if ret := C.tiledb_ctx_get_stats(c.tiledbContext, &stats); ret != C.TILEDB_OK {
-		return "", fmt.Errorf("Error getting stats from context: %s", c.LastError())
+		return nil, fmt.Errorf("Error getting stats from context: %s", c.LastError())
```
thetorpedodog, line 158: feel free to take or leave this idea, but you might consider returning `b'null'` (or `b'{}'`) here so that even in error cases the returned value is valid JSON.

thetorpedodog (approved):

context.go
snagles: let's proceed with this as is for now. I'm not opposed to it, but in this case I feel like if we return an error from this function the expectation would be that the json result does not contain any valid information (jmo)

PR #181, opened 2021-07-30
Fix segfault when attempting to DeserializeArraySchema invalid data.

thetorpedodog:
https://github.com/TileDB-Inc/TileDB/blob/ecfb3c629a8f3cf9377661afa94bedd0f26ec26f/tiledb/sm/c_api/tiledb.cc#L5275-L5290

In tiledb_deserialize_array_schema, `*array_schema` will be
unconditionally set to a newly-allocated address. If an error later
occurs, `*array_schema` remains at that value, but the memory that it
points to *has already been freed*. That means that, on the Go side,
it sees the `ArraySchema.tiledbArraySchema` as a non-nil pointer,
so when it goes to free it in `ArraySchema.Free`, the memory will be
freed again.

This change works around that by only adding the finalizer *after*
it is verified that there was no error.

---

Editor’s notes:

- It would be good to add tests on the other functions but I don’t know of a reliable way to get them to error out unlike trying to deserialize an invalid schema.
- We should probably set returned-but-freed out-param pointers to nullptr in the C API.
- It is well after midnight, do not be like me, go to sleep at a reasonable time.

### DIFF
--- a/serialize.go
+++ b/serialize.go
@@ -43,10 +43,6 @@ func SerializeArraySchema(schema *ArraySchema, serializationType SerializationTy
 // DeserializeArraySchema deserializes a new array schema from the given buffer
 func DeserializeArraySchema(buffer *Buffer, serializationType SerializationType, clientSide bool) (*ArraySchema, error) {
 	schema := ArraySchema{context: buffer.context}
-	// Set finalizer for free C pointer on gc
-	runtime.SetFinalizer(&schema, func(arraySchema *ArraySchema) {
-		arraySchema.Free()
-	})
 
 	var cClientSide C.int32_t
 	if clientSide {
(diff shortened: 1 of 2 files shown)


Shelnutt2 (changes requested):
This is actually caused by a defect in the c-api. I'll put a PR update in the TileDB core library today to fix it. This PR was super helpful to identify and highlight the defect.

Defect:
Not all error paths that call `delete` set the point back to `nullptr`.

thetorpedodog:
Could we merge this anyway in parallel? Being able to cherrypick this for the version of TileDB-Go we use on the server side will be helpful for working around a bug before we are able to fully upgrade our version of the core library.

Shelnutt2 (approved):
Accepting to unblock development :+1: , we'll get the core update and released in `2.3.3`

PR #182, opened 2021-07-30
Expand documentation of all x.Free() functions.

thetorpedodog:
This change replaces the doc comments for all the Free functions with
a more descriptive one that addresses that (a) it is idempotent and
(b) it is called automatically.

### DIFF
--- a/array.go
+++ b/array.go
@@ -68,7 +68,11 @@ func NewArray(tdbCtx *Context, uri string) (*Array, error) {
 	return &array, nil
 }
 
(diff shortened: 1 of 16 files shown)


snagles:
Maybe we should discuss unexporting these if it isn't necessary for the end user?

snagles (approved):

thetorpedodog:
That might be worth talking about in the future, but right now we manually free structures in the cloud server, so it does have at least *one* user.

PR #185, opened 2021-08-02
Update CI workflows to only cover supported versions of Go.

thetorpedodog:
Our go.mod file says we require Go 1.15, and now that Go 1.16 is out,
v1.14 is no longer supported.


Shelnutt2 (approved):

thetorpedodog:
This was a draft that was supposed to be only for me! (That said, if you’re approving, I’ll take it.)

PR #189, opened 2021-08-24
Support schema evolution

antalakas:


.github/workflows/tiledb-go.yml
```diff
@@ -121,12 +121,91 @@ jobs:
...
+      with:
+        go-version: ${{ matrix.go }}
+
+    - name: Cache dependencies
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.cache/go-build
+          ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
```
thetorpedodog, line 153: You should likely add -experimental at the end of this so that we don’t contaminate our non-experimental caches with our experimental caches.

array_schema_evolution.go
```diff
@@ -0,0 +1,111 @@
+//go:build experimental
+// +build experimental
+
+/**
+ * @file   array_schema_experimental.go
+ *
+ * @section DESCRIPTION
```
thetorpedodog, line 7: These Javadoc-like tags are not needed.

Also, traditional Go style is not to include the intervening `*`s in multiline comments.

array_schema_evolution.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// NewArraySchemaEvolution creates a TileDB schema evolution object.
+func NewArraySchemaEvolution(tdbCtx *Context) (*ArraySchemaEvolution, error) {
+	arraySchemaEvolution := ArraySchemaEvolution{context: tdbCtx}
+	ret := C.tiledb_array_schema_evolution_alloc(
+		arraySchemaEvolution.context.tiledbContext,
+		&arraySchemaEvolution.tiledbArraySchemaEvolution)
+	if ret != C.TILEDB_OK {
+		return nil, fmt.Errorf("Error creating tiledb arraySchemaEvolution: %s",
```
thetorpedodog, line 42: Error strings should always start with a lowercase letter.

array_schema_evolution.go
```diff
@@ -0,0 +1,111 @@
...
+		ase.context.tiledbContext, ase.tiledbArraySchemaEvolution,
+		attribute.tiledbAttribute)
+
+	name, err := attribute.Name()
+	if err != nil {
+		return fmt.Errorf("Cannot get name from attribute: %s",
+			ase.context.LastError())
+	}
+
+	if ret != C.TILEDB_OK {
```
thetorpedodog, line 74: This should probably get put right by the `ret :=` above.

array_schema_evolution.go
```diff
@@ -0,0 +1,111 @@
...
+
+// AddAttribute adds an attribute to an array schema evolution.
+func (ase *ArraySchemaEvolution) AddAttribute(attribute *Attribute) error {
+	ret := C.tiledb_array_schema_evolution_add_attribute(
+		ase.context.tiledbContext, ase.tiledbArraySchemaEvolution,
+		attribute.tiledbAttribute)
+
+	name, err := attribute.Name()
+	if err != nil {
+		return fmt.Errorf("Cannot get name from attribute: %s",
```
thetorpedodog, line 70: Since this is the result of a Go error, it should use the `%w` to wrap `err` rather than getting something from the context (which I think would be the wrong result, since that has no connection to the current context).

array_schema_evolution.go
```diff
@@ -0,0 +1,111 @@
...
+	// Set finalizer for free C pointer on gc
+	runtime.SetFinalizer(&arraySchemaEvolution,
+		func(arraySchemaEvolution *ArraySchemaEvolution) {
+			arraySchemaEvolution.Free()
+		})
+
+	return &arraySchemaEvolution, nil
+}
+
+// Free destroys an array schema evolution, freeing associated memory
```
thetorpedodog, line 55: I would copy the text from the other `Free` functions which has additional explanation.

array_schema_evolution_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const array_schema_evolution_name = "array_schema_evolution"
+
+func TestArraySchemaEvolution(t *testing.T) {
+	config, err := NewConfig()
+	assert.Nil(t, err)
```
thetorpedodog, line 17: require.NoError(t, err)

array_schema_evolution_test.go
```diff
@@ -0,0 +1,147 @@
...
+	err = arraySchema.SetCapacity(100)
+	assert.Nil(t, err)
+
+	err = arraySchema.SetDomain(domain)
+	assert.Nil(t, err)
+
+	err = arraySchema.Check()
+	assert.Nil(t, err)
+
+	array, err := NewArray(context, array_schema_evolution_name)
```
thetorpedodog, line 61: This array should probably go in a t.TempDir() rather than the current directory.

array_schema_evolution_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	err = array.Create(arraySchema)
+	assert.Nil(t, err)
+
+	err = array.Close()
+	assert.Nil(t, err)
+
+	arraySchemaEvolution, err := NewArraySchemaEvolution(context)
+	assert.Nil(t, err)
+	defer arraySchemaEvolution.Free()
```
thetorpedodog, line 75: Since the finalizer is automatically called by the GC, this should not be needed.

array_schema_evolution_test.go
```diff
@@ -0,0 +1,147 @@
...
+	assert.Nil(t, err)
+
+	// Need to get the evolved schema
+	arrAchema, err := arr.Schema()
+	assert.Nil(t, err)
+	assert.NotNil(t, arrAchema)
+
+	hasAttr, err := arrAchema.HasAttribute("a1")
+	assert.Nil(t, err)
+	assert.Equal(t, false, hasAttr)
```
thetorpedodog, line 115: assert.False

dimension.go
```diff
@@ -43,7 +43,7 @@ func NewDimension(context *Context, name string, datatype Datatype, domain inter
 	domainType := reflect.TypeOf(domain).Elem().Kind()
 	extentType := reflect.TypeOf(extent).Kind()
 	if extentType != domainType {
-		return nil, fmt.Errorf("Domaing and extent do not have the same data types. Domain: %s, Extent: %s", domainType.String(), extentType.String())
+		return nil, fmt.Errorf("Domain and extent do not have the same data types. Domain: %s, Extent: %s", domainType.String(), extentType.String())
```
thetorpedodog, line 46: while you're here, this should probably be lowercased. Also, a .String() call is not needed if you use %v.

examples/vacuum_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build !experimental
```
thetorpedodog, line 1: Is there any specific reason *not* to run this with experimental features on?

antalakas: yes, getting this result:

```
--- FAIL: ExampleVacuumSparseArray (0.08s)
got:
Estimated query size in bytes for attribute 'a': 12
Estimated query size in bytes for dimension 'd': 12
Cell (1) has data 1
Cell (2) has data 2
Cell (3) has data 3
Num of fragments after 2 writes before consolidate: 3
Num of fragments after consolidate: 4
Num of fragments after vacuum: 2
Estimated query size in bytes for attribute 'a': 12
Estimated query size in bytes for dimension 'd': 12
Cell (1) has data 1
Cell (2) has data 2
Cell (3) has data 3
want:
Estimated query size in bytes for attribute 'a': 12
Estimated query size in bytes for dimension 'd': 12
Cell (1) has data 1
Cell (2) has data 2
Cell (3) has data 3
Num of fragments after 2 writes before consolidate: 2
Num of fragments after consolidate: 3
Num of fragments after vacuum: 1
Estimated query size in bytes for attribute 'a': 12
Estimated query size in bytes for dimension 'd': 12
Cell (1) has data 1
Cell (2) has data 2
Cell (3) has data 3
```

thetorpedodog: Since it's not obvious why this code should not be run with experimental flags (i.e., it currently fails), I would add a comment (after the build tag) that explains why it is disabled in that situation. Eventually, we should be able to remove the not-experimental constraint.

.github/scripts/install_tiledb_source_linux.sh
```diff
@@ -0,0 +1,9 @@
+set -e -x
+git clone https://github.com/TileDB-Inc/TileDB
+cd TileDB
+git checkout 09e665cad9a4e3ed9c28ff3d100a218080fb18e7
```
thetorpedodog, line 4: If you want to follow along with the progress on that branch, you could make the first line:

```
git clone -b bd/ch9808-schema-evolution-serialization https://github.com/TileDB-Inc/TileDB.git
```

.github/scripts/install_tiledb_source_linux.sh
```diff
@@ -0,0 +1,9 @@
...
+mkdir build && cd build
+cmake -DTILEDB_VERBOSE=OFF -DTILEDB_S3=ON -DTILEDB_SERIALIZATION=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local ..
+make -j$(nproc)
+sudo make -C tiledb install
+sudo ldconfig
```
thetorpedodog, line 9: add closing newline?

thetorpedodog (approved):
three minor suggestions; otherwise this looks good!

PR #193, opened 2021-09-14
Removes not experimental tag from vacuum test

antalakas:
Issue with fragments is resolved in TileDB Core 2.4, test passes


thetorpedodog (approved):
Are there any other places where the !experimental tag needs to be removed? May be worth a follow-up.

PR #198, opened 2021-11-14
Removes *_with_key and open_at deprecated apis

antalakas:


snagles (approved):

examples_lib/encryption.go
```diff
@@ -9,10 +9,25 @@ import (
 // The 256-bit encryption key, stored as a string for convenience.
 const encryption_key = "0123456789abcdeF0123456789abcdeF"
 
-func createEncryptedArray(dir string) {
+func getContextConfiguredForEncryption() *tiledb.Context {
```
thetorpedodog, line 12: Since we now have a function to create a config from a dict, we could use that here.
Also maybe "build" instead of "get" for a more descriptive verb?

examples_lib/encryption.go
```diff
@@ -52,13 +67,13 @@ func createEncryptedArray(dir string) {
 	checkError(err)
 	defer array.Free()
 
-	err = array.CreateWithKey(schema, tiledb.TILEDB_AES_256_GCM, encryption_key)
+	//err = array.CreateWithKey(schema, tiledb.TILEDB_AES_256_GCM, encryption_key)
```
thetorpedodog, line 70: stray temporary comment

thetorpedodog (approved):
A couple minor things but looks good otherwise

examples_lib/encryption.go
antalakas: 👍

examples_lib/encryption.go
antalakas: Thanks, replaced with `NewContextFromMap`

PR #206, opened 2021-12-16
Support VFS Ls

antalakas:
Changes in Github Actions:

 - Use of `macos-11`
 - Use of TileDB Core `2.5.3` instead of `2.5.2`

### DIFF
--- a/vfs.go
+++ b/vfs.go
@@ -631,8 +631,8 @@ func vfsLs(path *C.cchar_t, data unsafe.Pointer) int32 {
 	return 1
 }
 
-// Ls returns list of folders and files in a path
-func (v *VFS) Ls(path string) ([]string, []string, error) {
+// List returns list of folders and files in a path
+func (v *VFS) List(path string) ([]string, []string, error) {
 	cpath := C.CString(path)
 	defer C.free(unsafe.Pointer(cpath))
 
--- a/vfs.go
+++ b/vfs.go
@@ -603,3 +603,50 @@ func (v *VFSfh) fetchAndSetSize() error {
 
 	return nil
 }
+
+// FolderData is a type encapsulating list of folders and files
+type FolderData struct {
+	Folders []string
+	Files   []string
+	Vfs     *VFS
+}
+
+//export vfsLs
+func vfsLs(path *C.cchar_t, data unsafe.Pointer) int32 {
+	folderData := pointer.Restore(data).(*FolderData)
+
+	uri := C.GoString(path)
+
+	isDir, err := folderData.Vfs.IsDir(uri)
+	if err != nil {
+		return 0
+	}
+
+	if isDir {
+		folderData.Folders = append(folderData.Folders, uri)
+	} else {
+		folderData.Files = append(folderData.Files, uri)
+	}
+
+	return 1
+}
+
+// Ls returns list of folders and files in a path
+func (v *VFS) Ls(path string) ([]string, []string, error) {
+	cpath := C.CString(path)
+	defer C.free(unsafe.Pointer(cpath))
+
+	folderData := FolderData{
+		Folders: []string{},
+		Files:   []string{},
+		Vfs:     v,
+	}
+	data := pointer.Save(&folderData)
+
+	ret := C._vfs_ls(v.context.tiledbContext, v.tiledbVFS, cpath, data)
+	if ret != C.TILEDB_OK {
+		return nil, nil, fmt.Errorf("error in getting path listing %s: %s", path, v.context.LastError())
+	}
+
+	return folderData.Folders, folderData.Files, nil
+}
--- a/vfs_test.go
+++ b/vfs_test.go
@@ -165,8 +165,8 @@ func TestVFSFH(t *testing.T) {
 	assert.ElementsMatch(t, b, bRead)
 }
 
-// TestVFSLs validates vfs LsDir operation is successful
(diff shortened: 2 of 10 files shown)


vfs.go
```diff
@@ -603,3 +603,50 @@ func (v *VFSfh) fetchAndSetSize() error {
...
+		folderData.Folders = append(folderData.Folders, uri)
+	} else {
+		folderData.Files = append(folderData.Files, uri)
+	}
+
+	return 1
+}
+
+// Ls returns list of folders and files in a path
+func (v *VFS) Ls(path string) ([]string, []string, error) {
```
thetorpedodog, line 635: Do we want this to be called `ls` for consistency with the TileDB C code, or do we want this to be named `List`? I am partial to `List` so that the name is less opaque.

vfs_test.go
```diff
@@ -174,3 +164,108 @@ func TestVFSFH(t *testing.T) {
...
+
+	context, err := NewContext(config)
+	require.NoError(t, err)
+
+	vfs, err := NewVFS(context, config)
+	require.NoError(t, err)
+
+	tmpPath := filepath.Join(t.TempDir(), "somedir")
+	tmpPath2 := filepath.Join(tmpPath, "somedir2")
+	tmpPath3 := filepath.Join(tmpPath, "somedir3")
```
thetorpedodog, line 181: Is the third one of these necessary? It seems like purely a duplicate of the "somedir2" case (which itself might be better renamed "subdir")

vfs_test.go
```diff
@@ -174,3 +164,108 @@ func TestVFSFH(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.True(t, isFile)
+
+	folderList, fileList, err := vfs.Ls(tmpPath)
+	require.NoError(t, err)
+	assert.Equal(t, 2, len(folderList))
+	assert.Equal(t, 3, len(fileList))
+
+	// Remove Files
+	require.NoError(t, vfs.RemoveFile(tmpFilePath))
```
thetorpedodog, line 251: Since we're not testing vfs.RemoveFile I would just let the `t.TempDir` cleanup handle deleting all these files itself.

vfs_test.go
```diff
@@ -174,3 +164,108 @@ func TestVFSFH(t *testing.T) {
...
+	require.NoError(t, vfs.RemoveFile(tmpFilePath2))
+	require.NoError(t, vfs.RemoveFile(tmpFilePath3))
+
+	// Remove directories
+	require.NoError(t, vfs.RemoveDir(tmpPath))
+}
+
+func createFile(t *testing.T, vfs *VFS, path string) {
```
thetorpedodog, line 259:
```suggestion
func createFile(t testing.TB, vfs *VFS, path string) {
t.Helper()
```

vfs_test.go
```diff
@@ -174,3 +164,108 @@ func TestVFSFH(t *testing.T) {
...
+	require.NoError(t, vfs.Touch(path))
+
+	fh, err := vfs.Open(path, TILEDB_VFS_WRITE)
+	require.NoError(t, err)
+
+	bytes := []byte{0, 1, 2}
+	require.NoError(t, vfs.Write(fh, bytes))
+
+	bytes2, err := vfs.Read(fh, 0, uint64(len(bytes)))
```
thetorpedodog, line 268: outBytes and inBytes for clarity?

vfs_test.go
```diff
@@ -174,3 +164,108 @@ func TestVFSFH(t *testing.T) {
...
+	require.NoError(t, err)
+	assert.True(t, isFile)
+
+	isFile, err = vfs.IsFile(tmpFilePath3)
+	require.NoError(t, err)
+	assert.True(t, isFile)
+
+	folderList, fileList, err := vfs.Ls(tmpPath)
+	require.NoError(t, err)
+	assert.Equal(t, 2, len(folderList))
```
thetorpedodog, line 247: assert.Len or maybe assert on the actual contents?

thetorpedodog (commented):
The code itself looks good; I have a few suggestions on the test which are not major issues but should help them be a bit more understandable.

vfs_test.go
antalakas: 👍

vfs_test.go
antalakas: 👍

vfs_test.go
antalakas: 👍

vfs_test.go
antalakas: 👍

vfs.go
antalakas: List works for me too

vfs_test.go
antalakas: subdir ok

vfs_test.go
```diff
@@ -174,3 +164,103 @@ func TestVFSFH(t *testing.T) {
 	require.NoError(t, r.Close())
 	assert.ElementsMatch(t, b, bRead)
 }
+
+// TestVFSLs validates vfs LsDir operation is successful
+func TestVFSLs(t *testing.T) {
```
thetorpedodog, line 169: nit: rename to `List` to match

thetorpedodog (approved):

vfs_test.go
antalakas: 👍

PR #209, opened 2022-01-18
Do not error for empty metadata value

snagles:
Removes error case for when a nil pointer is returned for a given metadata value because it is empty. Removal of the check necessitated getting back a zero value from GetValue. Empty values occur even though they're abnormal.

### DIFF
--- a/enums.go
+++ b/enums.go
@@ -255,127 +255,116 @@ func (d Datatype) MakeSlice(numElements uint64) (interface{}, unsafe.Pointer, er
 
 // GetValue gets value stored in a void pointer for this data type
 func (d Datatype) GetValue(valueNum uint, cvalue unsafe.Pointer) (interface{}, error) {
-	var value interface{}
 	switch d {
 	case TILEDB_INT8:
 		if valueNum > 1 {
 			tmpValue := make([]int8, valueNum)
-			tmpslice := (*[1 << 46]C.int8_t)(unsafe.Pointer(cvalue))[:valueNum:valueNum]
+			tmpslice := (*[1 << 46]C.int8_t)(cvalue)[:valueNum:valueNum]
 			for i, s := range tmpslice {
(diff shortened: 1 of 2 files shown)


enums.go
```diff
@@ -255,127 +255,155 @@ func (d Datatype) MakeSlice(numElements uint64) (interface{}, unsafe.Pointer, er
...
-			value = tmpValue
-		} else {
-			value = *(*float64)(unsafe.Pointer(cvalue))
+			return tmpValue, nil
 		}
+		return *(*float64)(cvalue), nil
 	case TILEDB_CHAR:
-		tmpslice := (*[1 << 46]C.char)(unsafe.Pointer(cvalue))[:valueNum:valueNum]
-		value = C.GoString(&tmpslice[0])[0:valueNum]
+		if cvalue == nil {
```
Shelnutt2, line 390: Need to also check for a length of 0 because a zero-length string could be non-null. Empty string `""` is valid.

enums.go
```diff
@@ -255,127 +255,155 @@ func (d Datatype) MakeSlice(numElements uint64) (interface{}, unsafe.Pointer, er
...
+			return "", nil
+		}
+		tmpslice := (*[1 << 46]C.char)(cvalue)[:valueNum:valueNum]
+		return C.GoString(&tmpslice[0])[0:valueNum], nil
 	case TILEDB_STRING_ASCII:
-		tmpslice := (*[1 << 46]C.char)(unsafe.Pointer(cvalue))[:valueNum:valueNum]
-		value = C.GoString(&tmpslice[0])[0:valueNum]
+		if cvalue == nil {
```
Shelnutt2, line 396: Need to also check for a length of 0 because a zero-length string could be non-null. Empty string `""` is valid.

Shelnutt2 (approved):

PR #217, opened 2022-04-20
Add serialization helper func, create copy of bytes to prevent gc

snagles:
Adds serialization helper funcs on a buffer. 

Handles:
- Trim of null terminated core buffers for JSON
- Maintains null terminator for capnp
- Creates copy of buffers to limit issues with Go GC and buffer collection


Shelnutt2 (approved):
Looks good, and one clarifying point for general awareness. It's not that capnp messages are null delimited, it's that they should be considered binary and if the last byte happens to be a null delimiter value that's "ok".

PR #218, opened 2022-05-03
Pulls `runtime.SetFinalizer` calls into a `freeOnGC` function.

thetorpedodog:
This centralizes all the calls to `runtime.SetFinalizer` into one place,
and documents exactly where that function should be called. This reduces
repetition and hopefully makes things less error-prone.


snagles:
lgtm, needs to be rebased it seems

PR #221, opened 2022-05-03
Make Buffer safer by holding onto the last-set byte slice.

thetorpedodog:
Because the Go GC doesn't currently move objects, we can make Buffer
safe(r) by holding onto a reference to the last-passed-in byte slice.


thetorpedodog:
I was checking for other places we might want to do this and it’s basically the same thing we already do in `query.buffers`:

https://github.com/TileDB-Inc/TileDB-Go/blob/bba32b44735ebd8181fbb12738873dc7bf86a5dd/query.go#L321

snagles (approved):

PR #226, opened 2022-06-27
Support filestore

anastasop:


snagles (changes requested):

filestore.go
```diff
@@ -0,0 +1,180 @@
...
+	tdbCtx    *Context // the tiledb context for all read operations
+	arrayURI  string   // the uri of the array
+	arraySize int64    // the size of the array as returned by GetSize
+	bytesRead int64    // the total bytes read so far. Used as an offset for read operations
+}
+
+// NewFileStoreReader returns a FileStoreReader for the array
+// ArrayURI is the uri of an existing array with the filestore schema (see SchemaCreate)
+func NewFileStoreReader(tdbCtx *Context, arrayURI string) (*FileStoreReader, error) {
+	arraySize, err := GetSize(tdbCtx, arrayURI)
```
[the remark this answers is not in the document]
Shelnutt2: Agreed, this will be further true for writes.

filestore.go
```diff
@@ -0,0 +1,180 @@
...
+	}, nil
+}
+
+// Size return the uncompressed size of the undelying array
+func (fs *FileStoreReader) Size() int64 {
+	return fs.arraySize
+}
+
+// Read satisfies io.Reader
+func (fs *FileStoreReader) Read(p []byte) (n int, err error) {
```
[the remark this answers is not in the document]
anastasop: Uses `p` because go io.Reader uses `p`

anastasop:
@Shelnutt2 @snagles 

This PR is 2 commits.

The first is a direct mapping of the core file store functions, the same way as was done in Py and Java. The names of the bindings are the exact core names (URIImport, SchemaCreate etc). We can provide higher level APIs but these should also be exposed to the user. I prefer the names to be the same as the core functions but this is a minor issue.

The second is a higher level API FileStoreReader. This implements an `io.Reader` for a file. Note that there is no `io.Writer`. This is because the core function
```
TILEDB_EXPORT int32_t tiledb_filestore_buffer_import(
    tiledb_ctx_t* ctx,
    const char* filestore_array_uri,
    void* buf,
    size_t size,
    tiledb_mime_type_t mime_type) TILEDB_NOEXCEPT;
```
always replaces content, it cannot append. We can implement a writer that uses a `bytes.Buffer`, loads everything in memory and then calls `buffer_import` but this hides from the user the fact that everything is in memory and there can be overflows.

filestore.go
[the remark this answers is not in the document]
anastasop: For a reader, i like the check here. It is the equivalent of `os.Open` An error `array does not exist` is not inappropriate here. But in a subsequent `read` it will be.

filestore.go
```diff
@@ -0,0 +1,180 @@
...
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error exporting file: %s", tdbCtx.LastError())
+	}
+
+	return nil
+}
+
+// BufferImport overwrites the contents of the filestore array with the contents of data
+// ArrayURI is the uri of an existing array with the filestore schema (see SchemaCreate)
+func BufferImport(tdbCtx *Context, arrayURI string, data []byte, mimeType FileStoreMimeType) error {
```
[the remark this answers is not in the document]
anastasop: This is a direct binding to the core method. The user owns the buffer not the library, so it is safe to export

filestore.go
```diff
@@ -0,0 +1,180 @@
...
+import (
+	"errors"
+	"fmt"
+	"io"
+	"runtime"
+	"unsafe"
+)
+
+// FileStoreReader is an io.Reader for filestore files
+type FileStoreReader struct {
```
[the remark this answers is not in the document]
anastasop: Chose `File`

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"runtime"
+	"unsafe"
+)
+
+// CreateFileSchemaFor returns a schema optimized for the storage of the named file.
+// An empty name returns a general schema suitable for any file.
+func CreateFileSchemaFor(tdbCtx *Context, name string) (*ArraySchema, error) {
```
snagles, line 29: maybe CreatedNamedFileSchema and CreateUnnamedFileSchema? Guard checks should probably just be in those functions and share the schema creation logic. Otherwise with this comment I don't see the point of CreateGeneralFileSchema

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error exporting file: %s", f.tdbCtx.LastError())
+	}
+
+	return nil
+}
+
+// Import replaces the contents of f with the contents of the named local file.
+// Assumes that the TileDB array exists and is of the appropriate schema.
+func (f *File) Import(name string, mimeType FileStoreMimeType) error {
```
snagles, line 99: I don't think this should be a method on the file itself. It seems like it should be FileImport() which creates the file struct, imports, and returns the file for use

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+	ret := C.tiledb_filestore_size(f.tdbCtx.tiledbContext, uri, &size)
+	if ret != C.TILEDB_OK {
+		return 0, fmt.Errorf("Error getting file size: %s", f.tdbCtx.LastError())
+	}
+
+	return int64(size), nil
+}
+
+// Export writes the contents of f to the named local file.
+func (f *File) Export(name string) error {
```
snagles, line 83: maybe this should be fileName or path since it is a full uri? Name implies just the relative name, not the full uri

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+	ret := C.tiledb_filestore_uri_import(f.tdbCtx.tiledbContext, arrayURI, fileURI, C.tiledb_mime_type_t(mimeType))
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error importing file: %s", f.tdbCtx.LastError())
+	}
+
+	return nil
+}
+
+// CreateAndImport creates the array, initializes it with the appropriate schema and imports the file.
+func (f *File) CreateAndImport(name string, mimeType FileStoreMimeType) error {
```
snagles, line 114: Would you mind explaining the thinking behind placing all of these functions of the file struct? 

My thinking is the same as above, I think this should be an exported single func that returns a file that is imported and ready for use. The file struct as it is currently written with a constructor doesn't really "mean anything" since you can't really do anything with the file until import/export/create and import are executed.

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+	arraySize int64    // the size of the array as returned by GetSize
+	bytesRead int64    // the total bytes read so far. Used as an offset for read operations
+
+	cArrayURI *C.char // a malloced char* C string for arrayURI. Cached here to avoid malloc on every read.
+}
+
+// NewFileReader returns a FileReader for the array
+// ArrayURI is the uri of an existing array with the filestore schema.
+func NewFileReader(tdbCtx *Context, arrayURI string) (*FileReader, error) {
+	arraySize, err := NewFile(tdbCtx, arrayURI).Size()
```
snagles, line 144: I think consensus was that this should not be calculated as part of the constructor because it could be an expensive operation and size isn't necessary until later. I think vfs.IsDir or IsFile or something along those lines are more prudent to check existance

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &FileReader{
+		tdbCtx:    tdbCtx,
+		arrayURI:  arrayURI,
+		arraySize: arraySize,
+		cArrayURI: C.CString(arrayURI),
```
snagles, line 153: this should have a finalizer calling close, leaves too much to the user. 

Better yet the c string should only be created where it is going to be actually used and can be gc'ed with a defer immediately after use

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+type FileReader struct {
+	tdbCtx    *Context // the tiledb context for all read operations
+	arrayURI  string   // the uri of the array
+	arraySize int64    // the size of the array as returned by GetSize
+	bytesRead int64    // the total bytes read so far. Used as an offset for read operations
+
+	cArrayURI *C.char // a malloced char* C string for arrayURI. Cached here to avoid malloc on every read.
+}
+
+// NewFileReader returns a FileReader for the array
```
snagles, line 141: There shouldn't need to be a separate FileReader type. The reader interface itself can be satisfied by the file type

filestore.go
```diff
@@ -0,0 +1,282 @@
...
+	if err := bufferExport(r.tdbCtx, r.cArrayURI, r.bytesRead, p); err != nil {
+		return 0, err
+	}
+	r.bytesRead += int64(len(p))
+
+	return len(p), nil
+}
+
+// FileWriter is an io.Writer for filestore files
+// FileWriter is an inmemory writer. Each Write stores the data in an internal bytes.Buffer.
```
snagles, line 191: Same logic for reader should be extended for the separate writer. This should not need a separate exported type

snagles: References should also be to io.WriteCloser rather than separate io.Writer and io.Closer

filestore_test.go
```diff
@@ -0,0 +1,269 @@
...
+		require.NoError(t, err)
+		require.Equal(t, sha1.New().Sum(fileData), sha1.New().Sum(sink.Bytes()))
+	})
+}
+
+func TestFileWriter(t *testing.T) {
+	t.Run("CreateSchema", func(t *testing.T) {
+		tempDir := t.TempDir()
+
+		fileURI := "testdata/VLDB17_TileDB_Page1.pdf"
```
snagles, line 182: test above look much simple with the tables, not sure these ones are as simple or easy to follow

snagles (changes requested):

filestore.go
```diff
@@ -0,0 +1,290 @@
...
+}
+
+// CreateFile creates a TileDB array at arrayURI, suitable to store a file of mimeType, and opens it for writing.
+// The returned File is an inmemory writer. Each Write stores the data in an internal bytes.Buffer.
+// The actual TileDB array creation happens on Close.
+func (fs *FileStore) CreateFile(arrayURI string, mimeType FileStoreMimeType) (*File, error) {
+	return &File{
+		tdbCtx:       fs.tdbCtx,
+		arrayURI:     arrayURI,
+		createSchema: true,
```
snagles, line 211: imo we should separate file creation vs file writing to reduce the extra logic introduced on this bool.

i.e
`filestore.CreateFile() *File`or `tiledb.CreateFile() *File` that create the schema

and then write file

filestore.go
```diff
@@ -0,0 +1,290 @@
...
+		mimeType:     mimeType,
+	}, nil
+}
+
+// OpenFileRead opens for reading the array at arrayURI, which should have a filestore schema.
+func (fs *FileStore) OpenFileToRead(arrayURI string) (*File, error) {
```
snagles, line 217: I think I'm still confused on the needs for these functions. We don't really open a file in read/write mode so I don't think we need to distinguish this in any way?

filestore.go
```diff
@@ -0,0 +1,290 @@
...
+func (fs *FileStore) OpenFileToWrite(arrayURI string) (*File, error) {
+	return &File{
+		tdbCtx:   fs.tdbCtx,
+		arrayURI: arrayURI,
+	}, nil
+}
+
+// NewArraySchemaForFile allocs a new ArraySchema optimized for the storage of file.
+// An empty path returns a general schema suitable for any file.
+func NewArraySchemaForFile(tdbCtx *Context, filePath string) (*ArraySchema, error) {
```
snagles, line 242: Does this need to be exposed? Do we think users will create file array schemas without using the files api as an entrypoint?

filestore.go
```diff
@@ -0,0 +1,290 @@
...
+	"unsafe"
+)
+
+// FileStore is the entry point for the filestore operations.
+type FileStore struct {
+	tdbCtx *Context
+}
+
+// NewFileStore creates a FileStore instance.
+func NewFileStore(tdbCtx *Context) *FileStore {
```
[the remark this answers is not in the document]
anastasop: No strong opinion. I will go with package level `ImportFile`, `ExportFile` `CreateFile`

filestore.go
anastasop: For completeness i think it must be exposed. The core API exposes it and all other binding we provide py,java,C# expose it also

snagles (approved):

snagles:
Updated looks good 👍

PR #231, opened 2022-07-15
Set up caching to speed up build and test.

thetorpedodog:
This change unifies the build scripts between Mac OS and Linux,
and caches build outputs. This means that future builds of the same
version don’t need to rerun the entire build process; they can use the
previously-built version.

We do a few things to accomplish this:

- Move compilation/installation into the `~/tiledb-core/install`
  directory. This allows us to cache it.
- Set appropriate `CGO_*FLAGS` variables to point to that directory.
  This includes the `rpath` directive, which embeds the full path
  to the dynamic library in the binary, so it knows it needs to look in
  `~/tiledb-core/install/lib` for `libtiledb.so`.
- Matrixes things out to reduce repetition.

This also fixes the ASAN step, which was never actually compiling.


teo-tsirpanis:
Closing because of conflicts, and because CI is pretty fast nowadays, owing to the use of prebuilt binaries.

PR #236, opened 2022-10-14
Associate array schema with domain in func ExampleNewArray(). Before …

philosopher-engineer:
…fix, This function fails to create an array with 'Error creating tiledb array: [TileDB::ArraySchema] Error: Array schema check failed; Domain not set'


snagles (approved):
lgtm

PR #238, opened 2022-10-21
Replace C.GoString with C.GoStringN where applicable

Shelnutt2:
This allows us to support nulls in the strings. TileDB does not use null delimiters for any data, but the conversion from "C" to "Go" was treating nulls as delimiters. This solves this by using the size give by the core.


antalakas (approved):
looks good to me

PR #242, opened 2022-11-02
[PoC] Use `tiledb_query_submit_async` in `Query.Submit()`.

teo-tsirpanis:
The way we submit queries from Go is like all the other languages, through the `tiledb_query_submit` function. The problem with this function is that it blocks the calling thread until the query completes, [hindering the scalability of Go's lightweight threading model](https://news.ycombinator.com/item?id=30041573#:~:text=if%20there%27s%20a%20CGO%20call%20(call%20to%20a%20C%20func)%20which%20uses%20I/O%20itself%20it%20will%20block%20an%20entire%20worker%20thread%20because%20the%20Go%20runtime%20isn%27t%20aware%20of%20any%20of%20it).

This PR fixes that. The implementation of `Query.Submit()` is changed to create a channel, call `tiledb_query_submit_async` with a callback that will send a value to the channel, and wait until this value is received. We still "block" but at the Go level, allowing the thread to be immediately reused. `Query.SubmitAsync()` was marked as deprecated; just calling `Submit()` should be a much better choice than checking for the status in a loop.

> **Warning** I am not 100% sure about this. This is the first time I am writing code in Go, and some tests are failing on my machine. It could be [SC-22770](https://app.shortcut.com/tiledb-inc/story/22770), or some other bug around `tiledb_query_submit_async`. This PR's main goal is to popularize the use of asynchronous code on languages that can easily take advantage of. Let's see what CI thinks.


query.go
```diff
@@ -2969,6 +2977,14 @@ func (q *Query) Finalize() error {
 	return nil
 }
 
+//export tiledb_go_query_submit_async_callback
+func tiledb_go_query_submit_async_callback(arg C.uintptr_t) {
+	handle := cgo.Handle(arg)
+	cont := handle.Value().(chan int)
+	handle.Delete()
+	cont <- 0
```
teo-tsirpanis, line 2985: [The documentation of `tiledb_query_submit_async`](https://tiledb-inc-tiledb.readthedocs-hosted.com/en/latest/c-api.html#c.tiledb_query_submit_async) says that "long-running callbacks should be dispatched to another thread". Is there a way to switch to a different thread with with certainty? Setting the channel in a new goroutine perhaps?

teo-tsirpanis:
Opened [SC-23194](https://app.shortcut.com/tiledb-inc/story/23194/improve-scalability-of-query-submit) to track this. CI times out at `ExampleReadingIncompleteArray`, looks like SC-22770.

teo-tsirpanis:
Closing, besides the aforementioned bug I saw that [async queries are not supported for remote arrays](https://github.com/TileDB-Inc/TileDB/blob/4726f9ea86097e0b1cf72b363de781f39d2dccc6/tiledb/sm/query/query.cc#L2753-L2756).

PR #248, opened 2022-12-02
Simplify memory handling in dimension.go with generics.

thetorpedodog:
`tiledb_dimension_get_domain` does not copy the bytes from the domain into the passed-in pointer; instead it uses the passed-in pointer as an out value, which it writes to point to the memory that it owns. This means that we can eliminate our allocating code. At the same time, we also use generics to reduce a bunch of duplication.

---

A little bit of stuff I figured is a good thing for a Friday fix-it.

### DIFF
--- a/dimension.go
+++ b/dimension.go
@@ -353,131 +353,53 @@ func (d *Dimension) Domain() (interface{}, error) {
 		return nil, err
 	}
 
-	var ret C.int32_t
-	var domain interface{}
 	switch datatype {
 	case TILEDB_INT8:
-		cdomain := C.malloc(2 * C.sizeof_int8_t)
-		defer C.free(cdomain)
-		tmpDomain := make([]int8, 2)
-		ret = C.tiledb_dimension_get_domain(d.context.tiledbContext, d.tiledbDimension, &cdomain)
-		tmpslice := (*[1 << 46]C.int8_t)(unsafe.Pointer(cdomain))[:2:2]
-		for i, s := range tmpslice {
-			tmpDomain[i] = int8(s)
-		}
-		domain = tmpDomain
+		return domainInternal[int8](d)
 	case TILEDB_INT16:
-		cdomain := C.malloc(2 * C.sizeof_int16_t)
-		defer C.free(cdomain)
-		tmpDomain := make([]int16, 2)
-		ret = C.tiledb_dimension_get_domain(d.context.tiledbContext, d.tiledbDimension, &cdomain)
-		tmpslice := (*[1 << 46]C.int16_t)(unsafe.Pointer(cdomain))[:2:2]
-		for i, s := range tmpslice {
-			tmpDomain[i] = int16(s)
-		}
-		domain = tmpDomain
+		return domainInternal[int16](d)
 	case TILEDB_INT32:
-		cdomain := C.malloc(2 * C.sizeof_int32_t)
-		defer C.free(cdomain)
-		tmpDomain := make([]int32, 2)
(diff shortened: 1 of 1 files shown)


thetorpedodog:
part of [sc-23938]

dimension.go
```diff
@@ -353,131 +353,53 @@ func (d *Dimension) Domain() (interface{}, error) {
...
+
+func domainInternal[T any](d *Dimension) ([]T, error) {
+	// tiledb_dimension_get_domain writes *a pointer to the memory it owns*
+	// into cDomain, so we need to ensure that the domain stays alive for
+	// the entire duration of this call.
+	defer runtime.KeepAlive(d)
+	var cDomain unsafe.Pointer
+	ret := C.tiledb_dimension_get_domain(d.context.tiledbContext, d.tiledbDimension, &cDomain)
 	if ret != C.TILEDB_OK {
 		return nil, fmt.Errorf("Error getting tiledb dimension's domain: %s", d.context.LastError())
```
NullHypothesis, line 399:
```suggestion
		return nil, fmt.Errorf("Error getting tiledb dimension's domain: %w", d.context.LastError())
```

dimension.go
```diff
@@ -486,75 +408,48 @@ func (d *Dimension) Extent() (interface{}, error) {
...
 	case TILEDB_STRING_ASCII:
-		extent = nil
-	default:
-		return nil, fmt.Errorf("Unrecognized extent type: %d", datatype)
-	}
-	if ret != C.TILEDB_OK {
-		return nil, fmt.Errorf("Error getting tiledb dimension's extent: %s", d.context.LastError())
+		return nil, nil
 	}
+	return nil, fmt.Errorf("Unrecognized extent type: %d", datatype)
```
NullHypothesis, line 438:
```suggestion
	return nil, fmt.Errorf("unrecognized extent type: %d", datatype)
```

dimension.go
```diff
@@ -353,131 +353,53 @@ func (d *Dimension) Domain() (interface{}, error) {
...
 		}
-		domain = tmpDomain
+		return []bool{asUints[0] != 0, asUints[1] != 0}, nil
 	case TILEDB_STRING_ASCII:
-		domain = nil
-	default:
-		return nil, fmt.Errorf("Unrecognized domain type: %d", datatype)
+		return nil, nil
 	}
+	return nil, fmt.Errorf("Unrecognized domain type: %d", datatype)
```
NullHypothesis, line 388:
```suggestion
	return nil, fmt.Errorf("unrecognized domain type: %d", datatype)
```

dimension.go
```diff
@@ -353,131 +353,53 @@ func (d *Dimension) Domain() (interface{}, error) {
...
+}
+
+func domainInternal[T any](d *Dimension) ([]T, error) {
+	// tiledb_dimension_get_domain writes *a pointer to the memory it owns*
+	// into cDomain, so we need to ensure that the domain stays alive for
```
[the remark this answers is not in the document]
thetorpedodog: It may be unnecessary in a practical sense, but purely in the sense of wanting this function to be correct on its own, I want to ensure that it is not GC'd.

More importantly, there's nothing special about the receiver object in methods; it is allowed to get GC'd after it is last referenced, just like any other variable or function parameter. I put together [a toy example on the Go Playground](https://go.dev/play/p/8w--oEScEBX), where `val` is finalized before `val.Method()` is completed.

thetorpedodog (commented):
Thank you! I fixed all the error messages you noted as well as the below discussion.

NullHypothesis (approved):

PR #259, opened 2023-05-01
Add support for dimension labels

shaunrd0:
This adds support for dimension labels. I needed this support to continue working on [SC-25205](https://app.shortcut.com/tiledb-inc/story/25205/core-rest-support-for-dimension-labels) (See blocking [SC-28383](https://app.shortcut.com/tiledb-inc/story/28383/add-c-apis-for-dimension-label-rest-support) for more detail).

I'm generally new to Go, so please point out any weirdness and I will address asap. Specifically looking for help or advice on this PR in the following areas:

- [X] The changes to query.go depend on new dimension label APIs, so builds without the experimental tag will fail. I'll do some digging for ideas online, but curious if there's already a preferred way to prevent this.
- [x]  There doesn't seem to be a Subarray object in TileDB-Go so I could not implement dimension label APIs that operate on Subarray. Should we add one in this PR or create a story for follow up?
- [x]  I struggled to implement tiledb_array_schema_set_dimension_label_tile_extent. I'll take another crack at it, but decided to move on for today so I could get working on adding tests for the other new APIs here.


shaunrd0:
@anastasop all done making changes we discussed over slack this morning. Once this is merged I can make a separate PR to add bindings for the 2 C APIs I added in core for REST support that were removed in [c2964c5](https://github.com/TileDB-Inc/TileDB-Go/pull/259/commits/c2964c5cde3918808c9e9d8f67422c122fec441c). Thanks again for taking a look and offering some Go advice for cleaning up experimental builds :pray:

anastasop:
Rebased to latest master to get TileDB 2.17 and the Subarray APIs

dimension_label.go
```diff
@@ -0,0 +1,405 @@
...
+	if extentType != dimType.ReflectKind() {
+		return fmt.Errorf("Dimension and extent do not have the same data types. Dimension: %s, Extent: %s",
+			dimType.ReflectKind(), extentType)
+	}
+
+	// Create extent void*
+	var cExtent unsafe.Pointer
+	switch dimType {
+	case TILEDB_INT8:
+		tmpExtent := extent.(int8)
```
thetorpedodog, line 214: There are a few things going on here which will be annoying to deal with:

- Since Go is garbage-collected, memory becomes eligible for freeing (and thus should be considered as already freed) after the last time they are referred to.
- `unsafe.Pointer` intentionally is opaque to the GC, so something referred to by an `unsafe.Pointer` does not count as actually there.

This means that in this case:

 1. `tmpExtent` gets space allocated (on the stack, in this case), and the value of `extent` is copied into it.
 2. We create an `unsafe.Pointer` to `tmpExtent`.
 3. `tmpExtent` is no longer referred to, thus making it eligible for collection.
 4. We pass that pointer out to TileDB.

(This might be something we do elsewhere in this package; if so we need to fix it there too.)

In this case we need to ensure `tmpExtent` stays around. So I think we need to do something like:

```go
var extentPtr any
defer runtime.KeepAlive(extentPtr)

// Here we can also reduce the amount of repetition by using a type-switch.
switch extent := extent.(type) {
case int8:
  // This ensures that `extent` stays alive.
  extentPtr = &extent
  // and since cExtent is a pointer to the same thing as extentPtr, it's OK.
  cExtent := unsafe.Pointer(&extent)
// ...
}
```

(This is annoying and fiddly.)

dimension_label.go
```diff
@@ -0,0 +1,405 @@
...
+	ret := C.tiledb_array_schema_set_dimension_label_tile_extent(a.context.tiledbContext, a.tiledbArraySchema,
+		cName, C.tiledb_datatype_t(dimType), cExtent)
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error setting dimension label tile extent on ArraySchema: %s", a.context.LastError())
+	}
+
+	return nil
+}
+
+// getDimensionLabelDataType Retrieve a dimension label Datatype from the schema using experimental APIs.
```
thetorpedodog, line 260: *retrieves a dimension label

enums.go
```diff
@@ -792,3 +792,39 @@ const (
...
+	ret := C.tiledb_data_order_to_str(C.tiledb_data_order_t(d), &dataOrderStr)
+	if ret != C.TILEDB_OK {
+		return "", fmt.Errorf("Error converting DataOrder to string: %d", d)
+	}
+
+	return C.GoString(dataOrderStr), nil
+}
+
+// FromString Converts from a string to the equivalent DataOrder enum
+func (d *DataOrder) FromString(name string) error {
```
thetorpedodog, line 820: This seems like an odd API:

```go
var myDO DataOrder
myDO.FromString("someStr")
```

I would expect something like:

```go
order, err := tiledb.DataOrderFromString(someStr)
```

dimension_label.go
```diff
@@ -0,0 +1,405 @@
...
+		return 0, fmt.Errorf("Error fetching dimension index for dimension label: %s", d.context.LastError())
+	}
+
+	return uint32(dimensionIndex), nil
+}
+
+// AttributeName returns the name of the attribute the label data is stored under.
+func (d *DimensionLabel) AttributeName() (string, error) {
+	var labelAttrName *C.char
+	ret := C.tiledb_dimension_label_get_label_attr_name(d.context.tiledbContext, d.tiledbDimensionLabel, &labelAttrName)
```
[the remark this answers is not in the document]
anastasop: Memory is owned by core but GoString at return makes a safe copy in the go space

dimension_label.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	return hasLabel != 0, nil
+}
+
+// DimensionLabelsNum returns the number of dimension label in this array schema
+func (a *ArraySchema) DimensionLabelsNum() (uint64, error) {
+	var labelNum uint64
+
+	ret := C.tiledb_array_schema_get_dimension_label_num(a.context.tiledbContext, a.tiledbArraySchema, (*C.uint64_t)(unsafe.Pointer(&labelNum)))
```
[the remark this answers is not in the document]
anastasop: Fixed

enums.go
anastasop: Correct. Made it a package method

dimension_label.go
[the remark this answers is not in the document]
thetorpedodog: Can you add a comment saying this? Just so that future readers know why this is safe. (Similarly in other cases where we get a reference to C-owned memory that we copy out of.)

examples_lib/utils.go
```diff
@@ -11,6 +11,11 @@ func checkError(err error) {
 	}
 }
 
+func checkedValue[T any](v T, err error) T {
```
thetorpedodog, line 14: could you add a docstring here? a one-liner should suffice.

dimension_label.go
shaunrd0: Done, thanks! Similar pattern was also in dimension.go, I made this change there too.

dimension_label.go
```diff
@@ -0,0 +1,408 @@
...
+		cExtent = unsafe.Pointer(&tmpExtent)
+	case float32:
+		extentPtr = &tmpExtent
+		cExtent = unsafe.Pointer(&tmpExtent)
+	case float64:
+		extentPtr = &tmpExtent
+		cExtent = unsafe.Pointer(&tmpExtent)
+	case bool:
+		extentPtr = &tmpExtent
+		cExtent = unsafe.Pointer(&tmpExtent)
```
thetorpedodog, line 248: i love go

thetorpedodog (approved):

PR #260, opened 2023-05-03
Sa/sc 28015/bindings for delete group

anastasop:
Add bindings for delete group https://tiledb-inc-tiledb.readthedocs-hosted.com/en/stable/c%2B%2B-api.html#_CPPv4N6tiledb5Group12delete_groupERKNSt6stringEb


anastasop:
@snagles draft until we get feedback if it is better to release this with core 2.15.3

snagles (approved):

PR #267, opened 2023-08-08
Support the TileDB core Subarray API for queries

anastasop:
This adds support for the TileDB core Subarray API for queries, see https://tiledb-inc-tiledb.readthedocs-hosted.com/en/stable/c-api.html#subarray

The API uses go generics to avoid casts and passing type unsafe `interface{}` values. Check the tests for usage examples.

### DIFF
--- a/range.go
+++ b/range.go
@@ -6,8 +6,8 @@ import (
 	"reflect"
 )
 
-// TileDBDimensionType is a constraint for the types allowed for a TileDB dimension
-type TileDBDimensionType interface {
+// DimensionType is a constraint for the types allowed for a TileDB dimension
+type DimensionType interface {
 	~string | ~float32 | ~float64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~int | ~int8 | ~int16 | ~int32 | ~int64
 }
 
@@ -18,14 +18,14 @@ type Range struct {
 }
 
 // MakeRange returns a typed range [from, to]. It can be used with AddRange to add ranges to a dimension.
-func MakeRange[T TileDBDimensionType](start, end T) Range {
+func MakeRange[T DimensionType](start, end T) Range {
 	return Range{start: start, end: end}
 }
 
 // ExtractRange extracts the endpoints of the range.
 // It returns []T{start, end, stride}. The stride is not supported by TileDB core yet,
 // so it gets the zero value of T
-func ExtractRange[T TileDBDimensionType](r Range) ([]T, error) {
+func ExtractRange[T DimensionType](r Range) ([]T, error) {
 	// we compare reflect.Kind because they give more versatility. Reflect.Type is more strict
 	// Consider:
 	// type Tag string
--- a/range.go
+++ b/range.go
@@ -1,0 +1,63 @@
+package tiledb
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+)
+
+// TileDBDimensionType is a constraint for the types allowed for a TileDB dimension
+type TileDBDimensionType interface {
+	~string | ~float32 | ~float64 | ~uint | ~uint8 | ~uint16 | ~uint32 | ~uint64 | ~int | ~int8 | ~int16 | ~int32 | ~int64
+}
+
+// Range is an 1D range along a subarray dimension
+type Range struct {
+	start interface{} // start of range, inclusive
+	end   interface{} // end of range, inclusive
+}
+
+// MakeRange returns a typed range [from, to]. It can be used with AddRange to add ranges to a dimension.
+func MakeRange[T TileDBDimensionType](start, end T) Range {
+	return Range{start: start, end: end}
+}
+
+// ExtractRange extracts the endpoints of the range.
+// It returns [3]T{start, end, stride}. The stride is not supported by TileDB core yet,
+// so it gets the zero value of T
+func ExtractRange[T TileDBDimensionType](r Range) ([]T, error) {
+	tKind := genericType[T]().Kind()
+	rKind := reflect.ValueOf(r.start).Kind()
+	if tKind != rKind {
+		return nil, fmt.Errorf("cannot extract a range of %s to a slice of %s", rKind, tKind)
+	}
+
+	res := make([]T, 3)
+	res[0] = r.start.(T)
+	res[1] = r.end.(T)
+	// res[2] is stride
+
+	return res, nil
+}
+
+// assertCompatibility checks that the datatype of an array dimension are the same as the range's.
+func (r Range) assertCompatibility(dimType Datatype, dimIsVar bool) error {
+	dKind := dimType.ReflectKind()
+	rKind := reflect.ValueOf(r.start).Kind()
+	rIsVar := rKind == reflect.String
+
+	if dimIsVar && dKind != reflect.Uint8 {
+		return errors.New("only []byte var dimensions are supported")
+	}
+	if dimIsVar && !rIsVar {
+		return errors.New("dimension is of variable size but range is not")
+	}
+	if !dimIsVar && rIsVar {
+		return errors.New("range is of variable size but dimension is not")
+	}
+	if !dimIsVar && dKind != rKind {
+		return fmt.Errorf("dimension and range types mismatch, range: %s dimension: %s", rKind, dKind)
+	}
+
+	return nil
+}
--- a/range.go
+++ b/range.go
@@ -26,6 +26,12 @@ func MakeRange[T TileDBDimensionType](start, end T) Range {
 // It returns []T{start, end, stride}. The stride is not supported by TileDB core yet,
 // so it gets the zero value of T
 func ExtractRange[T TileDBDimensionType](r Range) ([]T, error) {
+	// we compare reflect.Kind because they give more versatility. Reflect.Type is more strict
+	// Consider:
+	// type Tag string
+	// type Label string
+	// Label and Tag have different Type but same Kind. To interface with TileDB Core we use Kind
+	// which is reflected to the core dimensions types.
 	tKind := genericType[T]().Kind()
 	rKind := reflect.ValueOf(r.start).Kind()
 	if tKind != rKind {
--- a/range.go
+++ b/range.go
@@ -23,13 +23,13 @@ func MakeRange[T TileDBDimensionType](start, end T) Range {
 }
 
 // ExtractRange extracts the endpoints of the range.
-// It returns [3]T{start, end, stride}. The stride is not supported by TileDB core yet,
+// It returns []T{start, end, stride}. The stride is not supported by TileDB core yet,
 // so it gets the zero value of T
 func ExtractRange[T TileDBDimensionType](r Range) ([]T, error) {
 	tKind := genericType[T]().Kind()
 	rKind := reflect.ValueOf(r.start).Kind()
 	if tKind != rKind {
-		return nil, fmt.Errorf("cannot extract a range of %s to a slice of %s", rKind, tKind)
+		return nil, fmt.Errorf("cannot extract a range of %T to a slice of %v", r.start, genericType[T]())
 	}
 
 	res := make([]T, 3)
--- a/subarray.go
+++ b/subarray.go
@@ -87,12 +87,8 @@ func (sa *Subarray) AddRange(dimIdx uint32, r Range) error {
 		ret = C.tiledb_subarray_add_range_var(sa.context.tiledbContext, sa.subarray, C.uint32_t(dimIdx),
 			unsafe.Pointer(&startSlice[0]), C.uint64_t(len(startSlice)), unsafe.Pointer(&endSlice[0]), C.uint64_t(len(endSlice)))
 	} else {
-		sp := reflect.New(reflect.ValueOf(r.start).Type())
-		sp.Elem().Set(reflect.ValueOf(r.start))
-		ep := reflect.New(reflect.ValueOf(r.end).Type())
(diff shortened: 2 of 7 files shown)


range.go
```diff
@@ -0,0 +1,88 @@
+package tiledb
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+)
+
+// Range is an 1D range along a subarray dimension
+type Range struct {
```
thetorpedodog, line 10: Would it make sense to have the range type be generic? That means the user can’t accidentally unpack the wrong type of range.

anastasop: I would love to but go does not allow generic methods, cannot do `(sa *Subarray) methods[T](v T)`

thetorpedodog: You define methods on the generic type, so it would look like:

```go
func (sa *Subarray[T]) DoSomething() (T, error) {
  // ...
}
```

See [a complete example I banged together](https://go.dev/play/p/EFXirsf_3Yh) in the Go Playground.

anastasop: Correct. A subarray can have dimensions of different types. Should we go for types `Subarray1[T], Subarray2[T, V]`?

anastasop: `AddRange` is smooth and can be implemented with full static type checking. `GetRange` introduces an asymmetry, it cannot return by definition anything typed so it is up to the caller to use the proper `ExtractRange`. We can handle such issues if we rewrite `Array` with generics and keep track of the dimension types. This can be a followup.

The other thing is `Var` handling. True `Range` and `Range with var` are different. However if we implement them this way we will have types `Range, RangeVar` and methods `AddRange, AddRangeVar, GetRange, GetRangeVar` Now the methods are uniform, and we have one type of `Range` which we use in helpers.

range.go
```diff
@@ -0,0 +1,88 @@
...
+
+// MakeRange returns a variant range [from, to]. It can be used with AddRange to add ranges to a variable size dimension.
+func MakeRangeVar(start, end []byte) Range {
+	return Range{
+		kind:      reflect.Uint8,
+		isVar:     true,
+		startData: start,
+		endData:   end,
+	}
+}
```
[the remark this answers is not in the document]
anastasop: @thetorpedodog @snagles I separated the 2 types.

anastasop:
@snagles @thetorpedodog new implementation that unifies the `{Get/Add}Range` methods and simplifies the `var` dimensions

range.go
thetorpedodog: I now see the problem that you’re describing, since (as you note) the difficult part is getting a useful Range from the Subarray itself. I had not fully considered the implications of the roundtrip.

subarray.go
```diff
@@ -0,0 +1,263 @@
...
+	if isVar {
+		startSlice := []byte(r.start.(string))
+		endSlice := []byte(r.end.(string))
+		ret = C.tiledb_subarray_add_range_var_by_name(sa.context.tiledbContext, sa.subarray, cDimName,
+			unsafe.Pointer(&startSlice[0]), C.uint64_t(len(startSlice)), unsafe.Pointer(&endSlice[0]), C.uint64_t(len(endSlice)))
+	} else {
+		sp := reflect.New(reflect.ValueOf(r.start).Type())
+		sp.Elem().Set(reflect.ValueOf(r.start))
+		ep := reflect.New(reflect.ValueOf(r.end).Type())
+		ep.Elem().Set(reflect.ValueOf(r.end))
```
thetorpedodog, line 128: Apologies if I am missing something here, but could this be expressed as:

```go
sp := reflect.ValueOf(r.start).Addr()
// or maybe ....Elem().Addr() ?

File: subarray.go [hunk 4]
```diff
@@ -0,0 +1,263 @@
...
+			}
+		}
+	} else {
+		var startPointer, endPointer, stridePointer unsafe.Pointer
+		ret = C.tiledb_subarray_get_range(sa.context.tiledbContext, sa.subarray,
+			C.uint32_t(dimIdx), C.uint64_t(rangeNum), &startPointer, &endPointer, &stridePointer)
+		typ := dt.ReflectType()
+		if ret == C.TILEDB_OK {
+			r.start = reflect.NewAt(typ, startPointer).Elem().Interface()
+			r.end = reflect.NewAt(typ, endPointer).Elem().Interface()
```
thetorpedodog, line 204: It looks to me like this is going to give us a pointer into unmanaged memory. What guarantees do we have over the lifetime of the memory that is passed back to us? How do we ensure that it neither leaks nor is prematurely freed?

subarray.go
anastasop: This was my first try but i got
```
panic: reflect.Value.Addr of unaddressable value [recovered]
	panic: reflect.Value.Addr of unaddressable value
```

Looks like reflection cannot give you the address of a value, the equivalent of `var i int; ip := &i` The only way to make it work what this, create a pointer, copy the value and get back the pointer

subarray.go
anastasop: The pointer is in memory managed by TileDB core (unsafe as Go is concerned) but i use it only temporarily to copy the value.

```
                var startPointer, endPointer, stridePointer unsafe.Pointer
                ret = C.tiledb_subarray_get_range(sa.context.tiledbContext, sa.subarray,
                        C.uint32_t(dimIdx), C.uint64_t(rangeNum), &startPointer, &endPointer, &stridePointer)
		typ := dt.ReflectType()
                if ret == C.TILEDB_OK {
                        r.start = reflect.NewAt(typ, startPointer).Elem().Interface()
                        r.end = reflect.NewAt(typ, endPointer).Elem().Interface()
                }
```

I use the `startPointer etc` to get the range from TileDB and then i copy the value of that pointer reflectively (via `Elem`) The value is copied to Range and the pointers are not longer used, so no leaks

range.go
```diff
@@ -0,0 +1,63 @@
...
+	end   interface{} // end of range, inclusive
+}
+
+// MakeRange returns a typed range [from, to]. It can be used with AddRange to add ranges to a dimension.
+func MakeRange[T TileDBDimensionType](start, end T) Range {
+	return Range{start: start, end: end}
+}
+
+// ExtractRange extracts the endpoints of the range.
+// It returns [3]T{start, end, stride}. The stride is not supported by TileDB core yet,
```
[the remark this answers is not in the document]
anastasop: Corrected doc. The method returned a slice

range.go
```diff
@@ -0,0 +1,63 @@
...
+// so it gets the zero value of T
+func ExtractRange[T TileDBDimensionType](r Range) ([]T, error) {
+	tKind := genericType[T]().Kind()
```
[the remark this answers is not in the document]
anastasop: Type is more strict that Kind. For example

```
type Label string
type Tag string
```

these 2 have different `Type` but equal `Kind`

For TileDB i think we need the `Kind` to be more versatile

subarray.go
thetorpedodog: Maybe pulling this out into a function would make it easier to understand, something like:

```go
// addressableValue copies the value of `val` into an addressable location.
func addressableValue(val any) reflect.Value {
  valVal := reflect.ValueOf(val)
  pointable := reflect.New(valVal.Type())
  pointable.Elem().Set(valVal)
  return pointable
}
```

thetorpedodog (commented):
This looks pretty good overall; I just have a few things to make sure I understand and suggestions before it goes in.

subarray.go
```diff
@@ -0,0 +1,263 @@
...
+	if err != nil {
+		return err
+	}
+	if err := r.assertCompatibility(dt, isVar); err != nil {
+		return err
+	}
+
+	cDimName := C.CString(dimName)
+	defer C.free(unsafe.Pointer(cDimName))
+
```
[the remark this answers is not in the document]
anastasop: I can't see the path. `r` is passed as an argument and is not GCed until the method end. Moreover it is not longer used after we use it to set core range. If you can describe a GC path i will write a test and add KeepAlive if needed

subarray.go
```diff
@@ -0,0 +1,263 @@
...
+		endSlice := []byte(r.end.(string))
+		ret = C.tiledb_subarray_add_range_var_by_name(sa.context.tiledbContext, sa.subarray, cDimName,
+			unsafe.Pointer(&startSlice[0]), C.uint64_t(len(startSlice)), unsafe.Pointer(&endSlice[0]), C.uint64_t(len(endSlice)))
+	} else {
+		sp := reflect.New(reflect.ValueOf(r.start).Type())
+		sp.Elem().Set(reflect.ValueOf(r.start))
+		ep := reflect.New(reflect.ValueOf(r.end).Type())
+		ep.Elem().Set(reflect.ValueOf(r.end))
+		ret = C.tiledb_subarray_add_range_by_name(sa.context.tiledbContext, sa.subarray, cDimName,
+			sp.UnsafePointer(), ep.UnsafePointer(), nil)
```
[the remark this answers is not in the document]
anastasop: It is my understanding too that TileDB copies. The C++ [docs](https://tiledb-inc-tiledb.readthedocs-hosted.com/en/stable/c++-api.html#subarray) have example with templates that pass values. The C API uses pointers. If core does not copy then operations will fail. I will verify this with the core team

subarray.go
thetorpedodog: Understood. Thank you for explaining!

subarray.go
```diff
@@ -0,0 +1,255 @@
...
+
+	cDimName := C.CString(dimName)
+	defer C.free(unsafe.Pointer(cDimName))
+
+	var ret C.int32_t
+	if isVar {
+		startSlice := []byte(r.start.(string))
+		endSlice := []byte(r.end.(string))
+		ret = C.tiledb_subarray_add_range_var_by_name(sa.context.tiledbContext, sa.subarray, cDimName,
+			unsafe.Pointer(&startSlice[0]), C.uint64_t(len(startSlice)), unsafe.Pointer(&endSlice[0]), C.uint64_t(len(endSlice)))
```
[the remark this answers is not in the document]
anastasop: Added defensive keepalives

subarray.go
```diff
@@ -0,0 +1,255 @@
...
+	} else {
+		ret = C.tiledb_subarray_add_range_by_name(sa.context.tiledbContext, sa.subarray, cDimName,
+			addressableValue(r.start).UnsafePointer(), addressableValue(r.end).UnsafePointer(), nil)
```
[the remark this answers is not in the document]
anastasop: Added defensive keepalives. I think it is sufficient to keep alive `r`

subarray.go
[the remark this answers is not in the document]
anastasop: This is standard cgo. Just make a C string to pass to core. No need to copy first

subarray.go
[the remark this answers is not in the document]
thetorpedodog: But the new addressableValue values are completely independent of R; they are copies of the data from R. That is, we’re doing something vaguely like

```go
startPtr := new(typeof r.start)
*startPtr = r.start
unsafeStart := unsafe.Pointer(startPtr)
// from here on out, we no longer refer to `startPtr` so it is eligible for GC.
```
making startPtr a new pointer on the heap. If r.start were a pointer itself, the data pointed to by r would be preserved, but not necessarily the `startPtr` pointer to that same data. However, the data will *also* be preserved by keeping alive startPtr, without separately needing to keep r alive.

subarray.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+	if err := r.assertCompatibility(dt, isVar); err != nil {
+		return err
+	}
+
+	cDimName := C.CString(dimName)
+	defer C.free(unsafe.Pointer(cDimName))
+
+	var ret C.int32_t
+	var startSlice, endSlice []byte
```
thetorpedodog, line 121: these can move into `if isVar` and become `startSlice := []byte(...)`, and then the `runtime.KeepAlive(startSlice)` etc. can be at the end of the if statement.

thetorpedodog (commented):
OK, two more minor things but other than that I think this is all good apart from that! Thank you for tolerating me over this review process :)

subarray.go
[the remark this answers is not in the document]
anastasop: Made the change for extra safety. But i think i was OK. `r.start` is passed to `addressableValue` and the `reflect` package makes copies. This is in the call place
```
                ret = C.tiledb_subarray_add_range(sa.context.tiledbContext, sa.subarray, C.uint32_t(dimIdx),
                        addressableValue(r.start).UnsafePointer(), addressableValue(r.end).UnsafePointer(), nil)
```
so the Keepalive just after maybe is reduntant

anastasop:
@thetorpedodog can you give a final look?

subarray.go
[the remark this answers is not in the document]
thetorpedodog: The thing that needs to be kept alive isn't `r.start` and `r.end`; it’s the `addressableValue`s, because they are otherwise immediately forgotten about. Even though it's on the same line, what this looks like internally is

```go
startPtr := addressableValue(r.start)
startUnsafe := startPtr.UnsafePointer()
// Since startUnsafe is unmanaged, as of here, `startPtr` and its associated Value
// are never referred to again and thus eligible for collection.
ret = C.tiledb_whatever(..., startUnsafe, ...)
// This means that it is possible that this C call read or write to/from newly freed memory.
```

I admit it is *unlikely* that a GC cycle runs right there, but technically possible.

thetorpedodog (approved):
Whoops! I read this and I think I intended to respond but forgot to do that important part. I have one thing to clarify from a previous comment but apart from this concern everything looks good.

thetorpedodog (approved):

PR #276, opened 2023-10-19
Deduplicate cgo flags

ihnorton:
1. cgo concatenates all flags in a package:
> All the cgo CPPFLAGS and CFLAGS directives in a package are concatenated and used to compile C files in that package. (https://pkg.go.dev/cmd/cgo)

  So we can put them in a standalone file and remove duplication, which
  causes linker warnings on some platforms.

2. Note that `cgo_flags.go` must:
   - start with `package tiledb`
   - the cgo comments must be immediately followed by `import "C"`
     (or else they are ignored)

### DIFF
--- a/array.go
+++ b/array.go
@@ -1,8 +1,6 @@
 package tiledb
 
 /*
-
-
 #include <tiledb/tiledb.h>
 #include <stdlib.h>
 */
--- a/array.go
+++ b/array.go
@@ -1,8 +1,8 @@
 package tiledb
 
 /*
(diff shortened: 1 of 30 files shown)


array.go
```diff
@@ -1,8 +1,8 @@
 package tiledb
 
 /*
-#cgo LDFLAGS: -ltiledb
-#cgo linux LDFLAGS: -ldl
+
+
```
thetorpedodog, line 5: could you get rid of these blank lines?

PR #277, opened 2023-10-20
Wrap tiledb_handle_load_array_schema_request

davisp:
This wraps the new tiledb_handle_load_array_schema_request function in TileDB.

I was a bit unsure of where to put this. But based on the prior art for `DeserializeLoadEnumerationsRequest`, I stuck it in `array_schema.go`.

There's also at least one more of these coming for another handler though. So if you want me to create a new `handlers.go` or `cloud_handlers.go` or some such, I could also move the Enumeration handler and add the third all in one PR if that's preferable.

Also, I have no idea on Go style guides. I just know that this compiles. When I run `go fmt` locally it changes a lot of things that do not appear as they should changed so I've not run it and am just crossing my fingers I aped the `DeserializeLoadEnumerationsRequest` code closely enough.

### DIFF
--- a/array_schema.go
+++ b/array_schema.go
@@ -10,6 +10,7 @@ import "C"
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"unsafe"
 )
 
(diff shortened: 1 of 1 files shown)


davisp:
Some of those test failures look like version incompatibilities and I can't decipher the others. I'm no Go expert so feel free to poke me with a clue stick if I've missed anything obvious.

Also, I believe this is the first ever Go PR I've ever submitted which is very exciting.

ihnorton (commented):
LGTM but I'll leave for @anastasop to get a second set of eyes on the Go code, and merge.

array_schema.go
```diff
@@ -425,3 +426,22 @@ func (a *ArraySchema) Type() (ArrayType, error) {
 
 	return ArrayType(arrayType), nil
 }
+
+// Handle a LoadArraySchemaRequest. This is used by TileDB-Cloud
```
anastasop, line 430: The convention for go doc is to start with the entity name like `HandleLoadArraySchemaRequest .....`

array_schema.go
```diff
@@ -425,3 +426,22 @@ func (a *ArraySchema) Type() (ArrayType, error) {
...
+func HandleLoadArraySchemaRequest(array *Array, serializationType SerializationType, request *Buffer) (*Buffer, error) {
+	response, err := NewBuffer(array.context)
+	if err != nil {
+		return nil, fmt.Errorf("error deserializing load array schema request: %s", array.context.LastError())
```
anastasop, line 434: milsleading, there is no deserialization here. Use sth like `cannot allocate response buffer`

anastasop:
@davisp the PR is OK. The tests fail because it is build against a TileDB 2.17.4 build. To eliminate the test failures you can cherry-pick the first commits of https://github.com/TileDB-Inc/TileDB-Go/pull/279 or wait for a TileDB 2.18 release before resubmitting. No changes requests are needed but cannot be merged until TileDB 2.18 is released which will lead to TileDB-Go 0.24

anastasop (approved):

ypatia:
This work has been covered by this https://github.com/TileDB-Inc/TileDB-Go/pull/329 , closing as not needed anymore.

PR #278, opened 2023-10-26
Support empty enumerations and extensions

anastasop:
Support functionality for enumerations in TileDB 2.17.3

### DIFF
--- a/enumeration.go
+++ b/enumeration.go
@@ -310,6 +310,73 @@ func (e *Enumeration) Values() (interface{}, error) {
 	return strs, nil
 }
 
+// ExtendEnumeration extends an existing enumeration to add more values. The returned value should be
+// used with schema evolution to make changes persistent.
+func ExtendEnumeration[T EnumerationType](tdbCtx *Context, e *Enumeration, values []T) (*Enumeration, error) {
+	if len(values) == 0 {
+		return nil, fmt.Errorf("Error extending enumeration: empty values")
+	}
+
+	eName, err := e.Name()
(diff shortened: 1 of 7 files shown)


enumeration.go
```diff
@@ -302,6 +310,73 @@ func (e *Enumeration) Values() (interface{}, error) {
...
+		for _, v := range values {
+			dataSize += reflect.ValueOf(v).Len()
+		}
+		data := make([]byte, 0, dataSize)
+		offsets := make([]uint64, 0, len(values))
+		defer runtime.KeepAlive(data)
+		defer runtime.KeepAlive(offsets)
+		var currOffset uint64
+		for _, v := range values {
+			data = append(data, reflect.ValueOf(v).String()...)
```
thetorpedodog, line 351: I wasn’t sure if `.String` here would return the stringy value itself or the result of the value’s `.String` method, so I tested it myself: https://go.dev/play/p/MBOtM1EnwZr

This will in fact do The Right Thing here, i.e. putting the bytes of the string into the slice rather than the tostringy string.

Maybe worth putting a custom type like this into the test cases to affirmatively document this (and verify it)? This could be as simple as replacing a `[]string` addition with a `[]bogus` type.

thetorpedodog (approved):

enumeration.go
anastasop: @thetorpedodog Added test. Same code in in `newEnumeration`

thetorpedodog (approved):

PR #279, opened 2023-11-03
New delete fragments endpoints

anastasop:
This implements the 2 new endpoints `tiledb_array_delete_fragments_v2` and `tiledb_array_delete_fragments_list` for the upcoming TileDB core 2.18.0

### DIFF
--- a/array.go
+++ b/array.go
@@ -1080,6 +1080,7 @@ func (a *Array) Config() (*Config, error) {
 }
 
 // DeleteFragments delete the range of fragments from startTimestamp to endTimestamp. The array needs to be opened modify exclusive.
+// Deprecated: Use the module DeleteFragments
 func (a *Array) DeleteFragments(startTimestamp, endTimestamp uint64) error {
 	curi := C.CString(a.uri)
 	defer C.free(unsafe.Pointer(curi))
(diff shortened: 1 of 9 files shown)


anastasop:
@snagles @thetorpedodog This is the final PR to support TileDB 2.18 and tag 0.24.0

array.go
```diff
@@ -1092,3 +1093,39 @@ func (a *Array) DeleteFragments(startTimestamp, endTimestamp uint64) error {
...
+	defer C.free(unsafe.Pointer(curi))
+
+	var list []*C.char
+	for _, furi := range fragmentURIs {
+		cfuri := C.CString(furi)
+		defer C.free(unsafe.Pointer(cfuri))
+		list = append(list, cfuri)
+	}
+
+	ret := C.tiledb_array_delete_fragments_list(tdbCtx.tiledbContext, curi, (**C.char)(unsafe.Pointer(&list[0])), C.size_t(len(list)))
```
thetorpedodog, line 1123: This is probably something we should handle more broadly in the future, but you can use `unsafe.SliceData` to get the address of the start of a slice.

array_test.go
```diff
@@ -461,6 +461,128 @@ func TestArray_DeleteFragments(t *testing.T) {
...
+
+	testStarted := time.Now()
+	var fragmentCreatedAt []time.Time
+
+	context, err := NewContext(nil)
+	require.NoError(t, err)
+	err = array.Open(TILEDB_WRITE)
+	require.NoError(t, err)
+	for i := 1; i <= 10; i += 2 {
+		time.Sleep(100 * time.Millisecond)
```
thetorpedodog, line 481: can you add a note explaining why this is here?

array_test.go
```diff
@@ -461,6 +461,128 @@ func TestArray_DeleteFragments(t *testing.T) {
...
+	array, err := newTestArray(t)
+	require.NoError(t, err)
+
+	testStarted := time.Now()
+	var fragmentCreatedAt []time.Time
+
+	context, err := NewContext(nil)
+	require.NoError(t, err)
+	err = array.Open(TILEDB_WRITE)
+	require.NoError(t, err)
```
thetorpedodog, line 479: For situations where the function returns just an error, maybe `require.NoError(t, array.Open(TILEDB_WRITE))` etc.?

serialize.go
```diff
@@ -510,3 +510,25 @@ func DeserializeQueryAndArray(context *Context, buffer *Buffer, serializationTyp
 	runtime.KeepAlive(buffer)
 	return array, query, nil
 }
+
+// HandleArrayDeleteFragmentsTimestampsRequest is used by TileDB cloud to handle DeleteFragments with tiledb:// uris
+func HandleArrayDeleteFragmentsTimestampsRequest(context *Context, array *Array, buffer *Buffer, serializationType SerializationType) error {
+	ret := C.tiledb_handle_array_delete_fragments_timestamps_request(context.tiledbContext, array.tiledbArray,
```
thetorpedodog, line 516: do we need to `runtime.KeepAlive` the buffer or anything else here?

thetorpedodog (commented):
Overall looks good! I have a few minor suggestions and questions left.

array.go
anastasop: Yes that was the initial code but needed to revert because there are some go1.19 installations and 1.19 does not have `SliceData`

array_test.go
anastasop: local disk is too fast and i needed fragments to have some time distance between them

serialize.go
anastasop: added them for safety

thetorpedodog (approved):

PR #296, opened 2024-02-29
Add tar flag to fix corrupt macOS workflow.

NullHypothesis:
Our macOS job's tar often fails with the error message "Can't restore
time".  The following issue's detective work revealed that the problem
is caused by Apple's tar choking on extracting sparse files:
https://github.com/actions/runner-images/issues/2619

This commit adds the `-S` flag to tar, which fixes the problem.


vlasisPit (approved):

thetorpedodog:
what a strange and delightful error message

PR #300, opened 2024-03-15
Simplify workflow.

NullHypothesis:
My previous PR https://github.com/TileDB-Inc/TileDB-Go/pull/292 introduced a file that contains shared variables.  It's more idiomatic to put these variables directly into the workflow, using the `env` variable.


snagles (approved):

.github/workflows/tiledb-go.yml
```diff
@@ -7,6 +7,12 @@ on:
   pull_request:
     branches: [ master ]
 
+env:
+  # The version of TileDB to test against.
+  CORE_VERSION: "2.20.1"
```
Shelnutt2, line 12: Why change the names?

NullHypothesis: I find these names more descriptive than what I used before, and got inspired by Paul's unmerged PR https://github.com/TileDB-Inc/TileDB-Go/pull/231.

PR #301, opened 2024-03-21
Support and upgrade to TileDB 2.21.0.

NullHypothesis:
This PR adds support for the two new datatypes `TILEDB_GEOM_WKB` and `TILEDB_GEOM_WKT` that were added as part of the minor version bump to TileDB 2.21.0.

The PR also upgrades TileDB-Go to 2.21.0.

I _think_ that's all we need to support 2.21.0 but would be great to get a second pair of eyes on this.

### DIFF
--- a/enumeration_test.go
+++ b/enumeration_test.go
@@ -284,7 +284,6 @@ func TestEnumerationQueryCondition(t *testing.T) {
 		require.NoError(t, err)
 		err = rQuery.Submit()
 		require.NoError(t, err)
-		require.Contains(t, err.Error(), "Enumeration value not found")
 		require.NoError(t, array.Close())
 	})
 }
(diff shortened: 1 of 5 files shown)


enumeration_test.go
```diff
@@ -283,8 +283,7 @@ func TestEnumerationQueryCondition(t *testing.T) {
 		_, err = rQuery.SetDataBuffer("cols", colsBuffer)
 		require.NoError(t, err)
 		err = rQuery.Submit()
-		require.Error(t, err)
-		require.Contains(t, err.Error(), "Enumeration value not found")
+		require.NoError(t, err)
```
thetorpedodog, line 286: This seems like a meaningful change. What’s the reason for it?

NullHypothesis: Quoting the commit message in eeb4540:
> Core PR https://github.com/TileDB-Inc/TileDB/pull/4707 relaxed the `QueryCondition` requirement for enumeration values to be valid.  This means that where we previously expected an error, we now get no error.

thetorpedodog: Thanks; I was reading the review message and not the original commits.

thetorpedodog (approved):

snagles (approved):

PR #302, opened 2024-04-03
Remove `experimental` build tag.

NullHypothesis:
This PR removes all `experimental` build constraints. In addition, this PR adds the suffix "_experimental.go" to all files that make use of experimental TileDB-Core functionality. For posterity, here's the shell script I wrote to determine what files use experimental functionality:

```shell
#!/usr/bin/env bash

for file in $(find . -name "*.go"); do
	exp=0

	for func in $(grep -o 'tiledb_[a-z_]*(' "$file"); do
		# Is the function defined in a header file that has "experimental" in it?
		header_file=$(grep -lr "TILEDB_EXPORT.*$func" /usr/local/include/tiledb)
		if [[ $header_file == *"experimental"* ]]; then
				exp=1
				break
		fi
	done
	if [ "$exp" -eq 1 ]; then
		echo "$file contains experimental functions."
	else
		if [ $(grep -c '#include <tiledb/tiledb_experimental.h>' "$file") -gt 0 ]; then
			echo "$file includes experimental header but uses no experimental functions."
		fi
	fi	
done
```


snagles (approved):
I think we have some other references in .githun/workflows/tiledb-go.yml but otherwise LGTM pending that cleaned up

PR #306, opened 2024-04-24
Migrate away from deprecated TileDB C APIs.

teo-tsirpanis:
[SC-45987](https://app.shortcut.com/tiledb-inc/story/45987/tiledb-go-migrate-away-from-deprecated-tiledb-c-apis)

This PR updates TileDB-Go's usage of deprecated C APIs. The first commit migrates the `Group.GetMember***` functions to use `tiledb_group_get_member_by_(index|name)_v2`, while also adding code to work with TileDB string handles. The second commit deprecates `Query.SubmitAsync`, whose underlying C API has been deprecated since 2.15.

After that, the only deprecated C APIs used by TileDB-Go are inside deprecated Go APIs.

### DIFF
--- a/group_experimental.go
+++ b/group_experimental.go
@@ -274,42 +274,47 @@ func (g *Group) GetMemberCount() (uint64, error) {
 }
 
 func (g *Group) GetMemberFromIndex(index uint64) (string, string, ObjectTypeEnum, error) {
-	var curi *C.char
-	defer C.free(unsafe.Pointer(curi))
+	var curi *C.tiledb_string_t
 
-	var cname *C.char
-	defer C.free(unsafe.Pointer(cname))
+	var cname *C.tiledb_string_t
 
 	var objectTypeEnum C.tiledb_object_t
-	ret := C.tiledb_group_get_member_by_index(g.context.tiledbContext, g.group, C.uint64_t(index), &curi, &objectTypeEnum, &cname)
+	ret := C.tiledb_group_get_member_by_index_v2(g.context.tiledbContext, g.group, C.uint64_t(index), &curi, &objectTypeEnum, &cname)
 	if ret != C.TILEDB_OK {
 		return "", "", TILEDB_INVALID, fmt.Errorf("Error getting member by index for group: %s", g.context.LastError())
 	}
+	defer C.tiledb_string_free(&curi)
+	defer C.tiledb_string_free(&cname)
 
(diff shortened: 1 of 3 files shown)


group_experimental.go
```diff
@@ -274,42 +274,47 @@ func (g *Group) GetMemberCount() (uint64, error) {
...
-	defer C.free(unsafe.Pointer(cname))
+	var cname *C.tiledb_string_t
 
 	var objectTypeEnum C.tiledb_object_t
-	ret := C.tiledb_group_get_member_by_index(g.context.tiledbContext, g.group, C.uint64_t(index), &curi, &objectTypeEnum, &cname)
+	ret := C.tiledb_group_get_member_by_index_v2(g.context.tiledbContext, g.group, C.uint64_t(index), &curi, &objectTypeEnum, &cname)
 	if ret != C.TILEDB_OK {
 		return "", "", TILEDB_INVALID, fmt.Errorf("Error getting member by index for group: %s", g.context.LastError())
 	}
+	defer C.tiledb_string_free(&curi)
```
teo-tsirpanis, line 286: `tiledb_string_free` returns a result, which I don't know what to do with here.

teo-tsirpanis:
CI is green.

group_experimental.go
NullHypothesis: Looks like [it always returns `TILEDB_OK`](https://github.com/TileDB-Inc/TileDB/blob/5298c2252a7ebb34021439553be84b7e5a1bf3d3/tiledb/api/c_api/string/string_api.cc#L50-L55). I think it's safe to ignore the return value; especially since `tiledb_group_get_member_by_index_v2` succeeded at this point, which is what the caller cares about.

NullHypothesis (approved):
LGTM, thanks for this!

group_experimental.go
```diff
@@ -274,42 +274,47 @@ func (g *Group) GetMemberCount() (uint64, error) {
 }
 
 func (g *Group) GetMemberFromIndex(index uint64) (string, string, ObjectTypeEnum, error) {
-	var curi *C.char
-	defer C.free(unsafe.Pointer(curi))
```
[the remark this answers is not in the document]
teo-tsirpanis: I don't think so; [`free(null)` is a no-op](https://stackoverflow.com/a/6084233/4964734).

Issue #312, opened 2024-05-15
TileDB-GO started failing build against nightly libtiledb


jdblischak:
I run nightly tests against the dev branch of libtiledb in [jdblischak/centralized-tiledb-nightlies](https://github.com/jdblischak/centralized-tiledb-nightlies). TileDB-GO started failing yesterday.

The last passing build was [Monday night](https://github.com/jdblischak/centralized-tiledb-nightlies/actions/runs/9072666561) with libtiledb [62098dc](https://github.com/TileDB-Inc/TileDB/commit/62098dc9dc14a017fdfe0dec1610b05e64496e8a) and TileDB-GO b0fd1c949ecaba3f15c0d8b46393bec11976c90d.

On [Tuesday night (yesterday)](https://github.com/jdblischak/centralized-tiledb-nightlies/actions/runs/9088948929) it failed with libtiledb [42dbf36](https://github.com/TileDB-Inc/TileDB/commit/42dbf36a350931f8a3ee1abf9373be6d9eac1f69) and TileDB-GO b0fd1c949ecaba3f15c0d8b46393bec11976c90d (ie no change in TileDB-GO).

I happened to be running my nightlies yesterday in order to troubleshoot an unrelated problem, and TileDB-GO actually failed [earlier in the day yesterday](https://github.com/jdblischak/centralized-tiledb-nightlies/actions/runs/9068143688) too with libtiledb [6aca860](https://github.com/TileDB-Inc/TileDB/commit/6aca860d2fade1bf069d0870dc949d8e74dcd79a). So the error was likely introduced in [one of these five commits](https://github.com/TileDB-Inc/TileDB/compare/62098dc...6aca860).

The code I use to build and test TileDB-GO is [here](https://github.com/jdblischak/centralized-tiledb-nightlies/blob/3febc7b5b1f60a92a13c9258d894495883305390/.github/workflows/linux.yml#L558), but briefly it is:

```sh
go get -t .

# libtiledb.so is installed in ./install-libtiledb/lib
export CPATH=$(pwd)/install-libtiledb/include
export LIBRARY_PATH=$(pwd)/install-libtiledb/lib
export LD_LIBRARY_PATH=$(pwd)/install-libtiledb/lib

go test -v ./...
```

The lines with `FAIL` are:

```sh
FAIL	github.com/TileDB-Inc/TileDB-Go [build failed]
FAIL	github.com/TileDB-Inc/TileDB-Go/examples [build failed]
```

<details>
[...]

PR #313, opened 2024-05-15
Update tiledb_deserialize_array params.

shaunrd0:
This updates `tiledb_deserialize_array` to pass in the array URI as a parameter added in TileDB-Inc/TileDB#4961. Just getting a PR up in case we want to merge to fix nightly failures reported in #312.


NullHypothesis:
Closing because I ended up cherry-picking your work over at https://github.com/TileDB-Inc/TileDB-Go/pull/319. Thanks for the patch!

PR #321, opened 2024-06-13
Shorten datatype conversions in enums.go with generics.

thetorpedodog:
Yet another step on the road of consolidating things and making memory handling more straightforward. This stuff eventually helps lead to removing the deprecation warnings that show up on every compile, I promise.

### DIFF
--- a/common.go
+++ b/common.go
@@ -1,7 +1,6 @@
 package tiledb
 
 import (
-	"reflect"
 	"unsafe"
 )
 
@@ -17,6 +16,5 @@ type scalarType interface {
 
 // slicePtr gives you an unsafe pointer to the start of a slice.
 func slicePtr[T any](slc []T) unsafe.Pointer {
-	hdr := (*reflect.SliceHeader)(unsafe.Pointer(&slc))
-	return unsafe.Pointer(hdr.Data)
+	return unsafe.Pointer(unsafe.SliceData(slc))
 }
--- a/enums.go
+++ b/enums.go
@@ -13,6 +13,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"time"
 	"unsafe"
 )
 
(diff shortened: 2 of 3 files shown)


NullHypothesis (approved):
I _think_ this should be fine but I don't understand the original code in its entirety – especially the `(*[1 << 46]C.float)(cvalue)[:valueNum:valueNum]` expressions. I'd feel more comfortable with the change if enum.go was tested comprehensively. Either way, thanks for refactoring!

thetorpedodog:
> especially the `(*[1 << 46]C.float)(cvalue)[:valueNum:valueNum]` expressions

It took me a while to understand what was going on here too, so from inside out:

 0. `cvalue` is an `unsafe.Pointer` (which we already knew).
 1. `[1 << 46]C.float` is an arbitrarily large array type, so its value would be simply a massive block of memory that would ordinarily be passed by value. (For instance, a `[6]int64` is memory-layout–equivalent to `struct{A, B, C, D, E, F int64}`, so passing it to a function is not like in C where it looks like a pointer, but instead [copies the array into the function](https://go.dev/play/p/_OUOwqONkY3).)
 2. That means `*[1 << 46]C.float` is a pointer to the start of that array, or what we’re used to arrays looking like in C.
 3. Converting an `unsafe.Pointer` to any other pointer is allowed, so `(*[1 << 46]C.float)(cvalue)` takes the `cvalue` `unsafe.Pointer` and treats it as the pointer to the start of a huge array of `C.float`s. To make things easier to understand below, I’ll just call it `hugeArray`.
 4. `[:valueNum:valueNum]` is slicing syntax. Unlike in Python, the last thing is not a stride or something else; it’s instead the capacity of the slice. so `hugeArray[:valueNum:valueNum]` returns a slice whose `data` pointer is `cvalue` and whose `len` and `cap` are both `valueNum`.

So after a whole bunch of gymnastics, the result is a `valueNum`-long slice of `C.float`s at `cvalue`, which are then copied to `tmpValue`.

There are a few other places in the codebase that do something like this that I would also like to eliminate, and additionally do some other work to enhance memory safety.

PR #322, opened 2024-06-20
Add binding for tiledb_array_consolidate_fragments.

shaunrd0:
This adds the  `tiledb_array_consolidate_fragments` binding required for cloud fragment list consolidation.

CI will fail on this PR until https://github.com/TileDB-Inc/TileDB/pull/5135 is merged and backported to the core version in use by TileDB-Go.

### DIFF
--- a/array_experimental.go
+++ b/array_experimental.go
@@ -113,7 +113,7 @@ func (a *Array) ConsolidateFragments(config *Config, fragmentList []string) erro
 	list, freeMemory := cStringArray(fragmentList)
 	defer freeMemory()
 
-	ret := C.tiledb_array_consolidate_fragments(a.context.tiledbContext, curi, (**C.char)(unsafe.Pointer(&list[0])), C.uint64_t(len(list)), config.tiledbConfig)
+	ret := C.tiledb_array_consolidate_fragments(a.context.tiledbContext, curi, (**C.char)(slicePtr(list)), C.uint64_t(len(list)), config.tiledbConfig)
 	if ret != C.TILEDB_OK {
 		return fmt.Errorf("Error consolidating tiledb array fragment list: %s", a.context.LastError())
 	}
--- a/array_experimental.go
+++ b/array_experimental.go
@@ -113,7 +113,7 @@ func (a *Array) ConsolidateFragments(config *Config, fragmentList []string) erro
 	list, freeMemory := cStringArray(fragmentList)
 	defer freeMemory()
 
-	ret := C.tiledb_array_consolidate_fragments(a.context.tiledbContext, curi, (**C.char)(unsafe.Pointer(&list[0])), C.size_t(len(list)), config.tiledbConfig)
+	ret := C.tiledb_array_consolidate_fragments(a.context.tiledbContext, curi, (**C.char)(unsafe.Pointer(&list[0])), C.uint64_t(len(list)), config.tiledbConfig)
 	if ret != C.TILEDB_OK {
 		return fmt.Errorf("Error consolidating tiledb array fragment list: %s", a.context.LastError())
 	}
--- a/array_experimental.go
+++ b/array_experimental.go
@@ -6,7 +6,11 @@ package tiledb
 #include <stdlib.h>
 */
 import "C"
-import "fmt"
+import (
+	"fmt"
+	"runtime"
+	"unsafe"
+)
 
 // ConsolidationPlan is a consolidation plan for array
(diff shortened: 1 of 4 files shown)


shaunrd0:
I opened [SC-51140](https://app.shortcut.com/tiledb-inc/story/51140/enable-fragment-list-consolidation-test-in-tiledb-go) for follow up to enable the test added in this PR. The test fails on 2.24.X but passes on 2.25.0 with the fix in https://github.com/TileDB-Inc/TileDB/pull/5135.

This is ready for review, the test failure was the only thing keeping this PR in draft.

ypatia:
@NullHypothesis @thetorpedodog this is now ready for review. I don't feel yet confident in reviewing Go code :/

common.go
```diff
@@ -20,3 +24,18 @@ func slicePtr[T any](slc []T) unsafe.Pointer {
...
+	for i, str := range stringList {
+		list[i] = C.CString(str)
+	}
+
+	return list, func() {
+		for _, str := range list {
+			C.free(unsafe.Pointer(str))
+		}
+	}
+}
```
thetorpedodog, line 41: I like this a lot.

array_experimental.go
```diff
@@ -94,3 +98,26 @@ func (cp *ConsolidationPlan) DumpJSON() (string, error) {
...
+		return fmt.Errorf("Config must not be nil for Consolidate")
+	}
+
+	curi := C.CString(a.uri)
+	defer C.free(unsafe.Pointer(curi))
+
+	list, freeMemory := cStringArray(fragmentList)
+	defer freeMemory()
+
+	ret := C.tiledb_array_consolidate_fragments(a.context.tiledbContext, curi, (**C.char)(unsafe.Pointer(&list[0])), C.uint64_t(len(list)), config.tiledbConfig)
```
thetorpedodog, line 116: `(unsafe.Pointer(&list[0]))` can be `slicePtr(list)`.

array_experimental_test.go
```diff
@@ -9,8 +9,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestGetConsolidationPlan(t *testing.T) {
-	// Create an 1d array
+func createTestArray(t *testing.T) *Array {
+	// Create a 1d array
```
thetorpedodog, line 13: since the function `createTestArray` would be visible to all tests in this package, it might be worth making the name more specific, since the array created may not be suitable for all tests. maybe `create1DTestArray`?

array_experimental_test.go
```diff
@@ -131,3 +138,62 @@ func TestGetConsolidationPlan(t *testing.T) {
...
 }
+
+func TestConsolidateFragments(t *testing.T) {
+	// The test is skipped pending a core release for 2.25.0 that includes this fix:
+	// https://github.com/TileDB-Inc/TileDB/pull/5135
+	t.Skip("Skipping fragment list consolidation SC-51140")
+
+	array := createTestArray(t)
+
+	numFrags := uint32(5)
```
thetorpedodog, line 149: It looks like you’re doing this for the `assert.Equal` below…

array_experimental_test.go
```diff
@@ -131,3 +138,62 @@ func TestGetConsolidationPlan(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// Check that the new consolidated fragment was created.
+	err = fragmentInfo.Load()
+	require.NoError(t, err)
+	fragInfoNum, err = fragmentInfo.GetFragmentNum()
+	require.NoError(t, err)
+	fragToVacuumNum, err := fragmentInfo.GetToVacuumNum()
+	require.NoError(t, err)
+	require.Equal(t, numFrags, fragToVacuumNum)
```
thetorpedodog, line 184: …but instead, you can write `require.EqualValues(t, numFrags, fragToVacuumNum)` (or `require.Equal(t, uint32(numFrags), ...)`) to avoid having a bunch of `uint32(n)`s above.

thetorpedodog (approved):
overall very good. a few minor suggestions that you can clean up at your leisure

array_experimental.go
shaunrd0: Done

array_experimental_test.go
shaunrd0: Done

array_experimental_test.go
shaunrd0: Much better, very nice :+1:

PR #325, opened 2024-07-02
Update Go versions to test.

NullHypothesis:
We should include the latest minor release (1.22) and we might as well stop testing 1.20.


NullHypothesis:
Closing and re-opening because of stuck workers.

PR #327, opened 2024-07-09
Remove deprecated APIs.

teo-tsirpanis:
Part of SC-50038

[SC-50561](https://app.shortcut.com/tiledb-inc/story/50561/tiledb-go-tests-are-using-deprecated-apis)

Consistent with the Core's imminent removal of deprecated APIs in TileDB-Inc/TileDB#5146, this PR removes all deprecated APIs in TileDB-Go, which are the only consumers of deprecated C APIs.

One API, `Query.GetRanges` was not removed its `Subarray.GetRanges` successor was just added and is not an exact replacement. I reimplemented it however on top of the `Subarray` range methods.

I also migrated all usages of deprecated APIs in tests and examples (the bulk of this PR's changes).

Validated by successfully building the REST server with this branch of TileDB-Go.


NullHypothesis:
Thanks for this, @teo-tsirpanis! There are [test errors](https://gitlab.com/TileDB-Inc/TileDB-REST-Prototype/-/pipelines/1370106947) when running REST's CI on your PR. Looks like most of these errors are caused by REST's continued use of the now-removed `SetSubArray` function, like [here](https://github.com/TileDB-Inc/TileDB-Cloud-REST/blob/3f28c2df41872c4fc1552d252ee729dbce36ba64/internal/platform/config/unit_test/array.go#L163-L168). I'm going to have to figure out how to rewrite these tests.

array_experimental_test.go
```diff
@@ -69,11 +69,17 @@ func TestGetConsolidationPlan(t *testing.T) {
 	err = array.Open(TILEDB_WRITE)
 	require.NoError(t, err)
 
+	// Create subarray
+	subarray, err := array.NewSubarray()
```
snagles, line 73: from a usability perspective it seems like NewSubarray could accept accept the parameters SetSubarray does and do the same logic. Do we ever SetSubarray on the same alloced subarray with different params?

subarray.go
```diff
@@ -52,6 +52,101 @@ func (sa *Subarray) Free() {
 	}
 }
 
+// SetSubarray sets a subarray, defined in the order dimensions were added.
```
NullHypothesis, line 55: Should be `SetSubArray` instead of `SetSubarray`, otherwise Go's doc tool may not pick up on this.

subarray.go
```diff
@@ -52,6 +52,101 @@ func (sa *Subarray) Free() {
...
+
+	if reflect.TypeOf(subArray).Kind() != reflect.Slice {
+		return fmt.Errorf("Subarray passed must be a slice, type passed was: %s", reflect.TypeOf(subArray).Kind().String())
+	}
+
+	subArrayType := reflect.TypeOf(subArray).Elem().Kind()
+
+	schema, err := sa.array.Schema()
+	if err != nil {
+		return fmt.Errorf("Could not get array schema from query array: %s", err)
```
NullHypothesis, line 68: We may have to update the error message because the method no longer operates on queries, no?

subarray.go
```diff
@@ -52,6 +52,101 @@ func (sa *Subarray) Free() {
...
+		// Create subArray void*
+		tmpSubArray := subArray.([]bool)
+		csubArray = unsafe.Pointer(&tmpSubArray[0])
+	default:
+		return fmt.Errorf("Unrecognized subArray type passed: %s", subArrayType.String())
+	}
+
+	ret := C.tiledb_subarray_set_subarray(sa.context.tiledbContext, sa.subarray, csubArray)
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error setting query subarray: %s", sa.context.LastError())
```
NullHypothesis, line 145: Same as above.

subarray.go
```diff
@@ -261,6 +261,69 @@ func (sa *Subarray) GetRangeNumFromName(dimName string) (uint64, error) {
 	return rangeNum, nil
 }
 
+// GetRanges gets the number of dimensions from the array under current query
```
NullHypothesis, line 264: Doc string may need an update because queries aren't involved here.

NullHypothesis (commented):
Thanks for tackling this significant task! Left a few minor comments.

array_experimental_test.go
teo-tsirpanis: Calling `SetSubArray` is one of the two options to configure a subarray object; the other is by calling `AddRange(ByName)` for each dimension.

`NewSubarray` could accept the raw values optionally as a shortcut (I don't like it personally but I don't have enough experience to have an opinion on Go API design), but we should better do it on a subsequent PR.

teo-tsirpanis:
Feedback addressed @NullHypothesis.

NullHypothesis (approved):

NullHypothesis:
Looks good now, thanks Teo. I understand that merging this PR is a pressing matter for core but considering the PR's complexity, I want to give others an opportunity to take a look. My plan is to merge by the end of today.

snagles (approved):

PR #329, opened 2024-07-15
Wrap tiledb_handle_load_array_schema_request.

shaunrd0:
This wraps tiledb_handle_load_array_schema_request added in https://github.com/TileDB-Inc/TileDB/pull/4399. This binding is required for https://github.com/TileDB-Inc/TileDB-Cloud-REST/pull/4686.


serialize.go
```diff
@@ -514,6 +514,23 @@ func DeserializeQueryAndArray(context *Context, buffer *Buffer, serializationTyp
 	return array, query, nil
 }
 
+// HandleLoadArraySchemaRequest Pass the array and serialized request to core which returns the serialized response.
```
thetorpedodog, line 517: Could this be made more descriptive? Most functions in this package pass something to core, so that doesn’t really illuminate much.

serialize.go
```diff
@@ -514,6 +514,23 @@ func DeserializeQueryAndArray(context *Context, buffer *Buffer, serializationTyp
...
+func HandleLoadArraySchemaRequest(array *Array, serializationType SerializationType, request *Buffer) (*Buffer, error) {
```
thetorpedodog, line 518: Looking at a few of our functions above, most of them seem to put the request in front of the serialization type in the parameter list.

serialize.go
shaunrd0: Done, thanks!

serialize.go
shaunrd0: Done

serialize.go
```diff
@@ -514,6 +514,25 @@ func DeserializeQueryAndArray(context *Context, buffer *Buffer, serializationTyp
...
+		return nil, fmt.Errorf("error creating LoadArraySchemaResponse buffer: %s", array.context.LastError())
+	}
+
+	ret := C.tiledb_handle_load_array_schema_request(array.context.tiledbContext, array.tiledbArray, C.tiledb_serialization_type_t(serializationType), request.tiledbBuffer, response.tiledbBuffer)
+	if ret != C.TILEDB_OK {
+		return nil, fmt.Errorf("error handling LoadArraySchemaRequset: %s", array.context.LastError())
+	}
+
+	runtime.KeepAlive(request)
+	runtime.KeepAlive(response)
```
thetorpedodog, line 532: This really isn't necessary, since the buffer is returned, but isn't harmful either.

thetorpedodog (approved):

serialize.go
shaunrd0: Removed :+1:

PR #334, opened 2024-07-16
Reapply simplification of memory handling with generics.

thetorpedodog:
This reverts #331 and #330, thus reintroducing #248 and #321.

The issue in those changes was that when attempting to copy a null/zero-length string, we would do something like `C.GoStringN(nil, 0)` or maybe `C.GoStringN(someInvalidAddress, 0)`. Either of those should (in theory) give you a zero-length string, but in practice we want to manually check (as the code did before I removed that check). The check remained present in the *other* uses, since they all had an equivalent check before their copying operation in `getValueInternal`.

This change is also currently running through a server-side test suite.

### DIFF
--- a/array_test.go
+++ b/array_test.go
@@ -400,6 +400,21 @@ func TestArray_Metadata(t *testing.T) {
 		a.Close()
 		assert.NoError(t, err)
 	})
+
+	t.Run("empty", func(t *testing.T) {
+		a, err := newTestArray(t)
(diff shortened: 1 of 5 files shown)


thetorpedodog:
The test I just added fails without the fix (i.e., commit 2a546a805547830a612a4ce2b7226806d071c64c) and passes with it.

snagles (approved):

PR #326, opened 2024-07-05
Update Go versions to test.

NullHypothesis:
We should include the latest minor release (1.22) and I see no harm in discontinuing testing with 1.20.

PR #336, opened 2024-07-24
Update Go versions to test.

NullHypothesis:
We should include the latest minor release (1.22) and I see no harm in discontinuing testing with 1.20.

This PR is identical to https://github.com/TileDB-Inc/TileDB-Go/pull/326 but I couldn't merge it because there are two now-obsolete jobs that are stuck, so I'm opening a new PR.


snagles (approved):

thetorpedodog:
If we’re doing this, perhaps we should update the `go.mod` file to say "go 1.21" as well.

NullHypothesis:
> If we’re doing this, perhaps we should update the `go.mod` file to say "go 1.21" as well.

On second thought, I think it's best to add 1.22 to the test array while keeping 1.20 in go.mod. After all, we're not using any new language features, so there's no strong reason to update.

thetorpedodog (approved):

PR #338, opened 2024-07-26
Improve and speed up CI.

NullHypothesis:
This PR makes the following changes:
* Fix the linter, address its warnings, and stop letting it fail.
* Fix the address sanitizer and stop letting it fail.
* Make it so the address sanitizer is only ran after a new release. It's an expensive job to run and there's no need to run it for each PR.


NullHypothesis:
@ypatia, @shaunrd0, @thetorpedodog, @snagles: A reminder about this PR. Most of it are minor changes to please the linter, so it shouldn't take too long to review.

.github/workflows/tiledb-go.yml
```diff
@@ -168,4 +167,4 @@ jobs:
     # Tests TileDB-Go
     - name: Running examples using address sanitizer flags
       continue-on-error: true
```
shaunrd0, line 169: This is intentionally still allowed to fail? Just double checking - I still have a story open to look into the errors from ASAN. If this is just allowed to fail until that story is completed I'll make sure to remove this when I'm finished so it's required to pass moving forward.

shaunrd0 (approved):
Just a quick question, LGTM

.github/workflows/tiledb-go.yml
NullHypothesis: Right, that's the main reason why it's allowed to fail. That said, I'm also fine with removing it from CI but keeping the script, so it can be run manually. In fact, I would prefer for it to be a CI job that must be triggered manually but GitHub doesn't seem to support that.

shaunrd0: Thanks! No need to change this PR unless you feel strongly, but I think if the ASAN job was moved to a separate workflow file we could use [workflow_dispatch](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually) event to trigger it manually. I don't think you can trigger a specific job manually though, just workflows.

NullHypothesis: > No need to change this PR unless you feel strongly, but I think if the ASAN job was moved to a separate workflow file we could use [workflow_dispatch](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually) event to trigger it manually.

Ah, good to know! For now, I suggest to merge the PR and revisit the topic once we took a look at the ASAN warnings.

PR #341, opened 2024-08-07
Allow nil contexts to be passed when they can be inferred.

thetorpedodog:
This makes it a bit more convenient to do a few things:

- You can now allocate a new Array and provide a nil Context and a new (empty) Context will be created for you.
- When you create a new Query from an Array, you can provide a nil context and the array’s context will be used.

---

Another thing I ran into while working on stuff, which might make things a little more convenient for the majority of these constructor functions.

### DIFF
--- a/array.go
+++ b/array.go
@@ -54,7 +54,15 @@ type NonEmptyDomain struct {
 }
 
 // NewArray allocates a new array.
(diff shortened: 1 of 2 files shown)


NullHypothesis (approved):
A nice quality-of-life improvement!

shaunrd0 (approved):

PR #342, opened 2024-08-14
Remove unnecessary `Sleep` calls in tests.

NullHypothesis:
As far as I can tell, these calls are not necessary and slow down our tests considerably.  After removing the `Sleep` calls, I ran the test multiple times as follows:

    go test \
        -race \
        -count=1 \
        -timeout 30s \
        -tags experimental \
        -run ^ExampleTimestampArray$ github.com/TileDB-Inc/TileDB-Go/examples

All tests passed, so we should be fine.

### DIFF
--- a/examples_lib/reading_timestamp.go
+++ b/examples_lib/reading_timestamp.go
@@ -241,15 +241,12 @@ func RunTimestampArray() {
 	// Write data and metadata
 	t1 := getTimestamp()
 	writeTimestampArray(tmpDir1, "meta_key", "Write1", t1, 0)
-	time.Sleep(2000 * time.Millisecond)
 	// Write metadata only
 	t2 := getTimestamp()
 	writeTimestampArrayMeta(tmpDir1, "meta_key", "Write2", t2)
-	time.Sleep(2000 * time.Millisecond)
 	// Write metadata only
 	t3 := getTimestamp()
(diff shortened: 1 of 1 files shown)


thetorpedodog (commented):
I think we still might want these to make sure these don’t land on the same millisecond (or at least to make it clear that they shouldn’t). Maybe reduce it to something like `5 * Millisecond` instead?

NullHypothesis:
> I think we still might want these to make sure these don’t land on the same millisecond (or at least to make it clear that they shouldn’t). Maybe reduce it to something like `5 * Millisecond` instead?

Right, I followed your suggestion in the follow-up commits.

examples_lib/reading_timestamp.go
```diff
@@ -241,15 +241,17 @@ func RunTimestampArray() {
 	// Write data and metadata
 	t1 := getTimestamp()
 	writeTimestampArray(tmpDir1, "meta_key", "Write1", t1, 0)
-	time.Sleep(2000 * time.Millisecond)
+	// Wait a few milliseconds to ensure that we are not going to overwrite
+	// data.
```
thetorpedodog, line 245: instead of "we are not going to overwrite the data" (which we want to do), it might be better to say "our writes are in different milliseconds to ensure correct ordering".

thetorpedodog (approved):
one minor suggestion on the comment but looks good!

PR #345, opened 2024-08-21
Fix ASAN failures.

shaunrd0:
This fixes ASAN failures seen on CI, it looks like they were due to missing calls to `Free()` either in the example code or internal definitions.

If this looks correct I can follow up and look for similar errors. At the moment this only fixes those reported by `build_with_sanitizer_and_run.sh` so ASAN CI should pass now, but I would expect there's other similar errors that are not covered by the examples.

### DIFF
--- a/.github/workflows/tiledb-go.yml
+++ b/.github/workflows/tiledb-go.yml
@@ -166,5 +166,4 @@ jobs:
 
     # Tests TileDB-Go
     - name: Running examples using address sanitizer flags
-      continue-on-error: true
       run: ./.github/scripts/build_with_sanitizer_and_run.sh
--- a/array.go
+++ b/array.go
@@ -375,11 +375,13 @@ func (a *Array) NonEmptyDomain() ([]NonEmptyDomain, bool, error) {
 	if err != nil {
 		return nil, false, err
 	}
+	defer schema.Free()
 
 	domain, err := schema.Domain()
 	if err != nil {
 		return nil, false, err
 	}
+	defer domain.Free()
 
 	ndims, err := domain.NDim()
 	if err != nil {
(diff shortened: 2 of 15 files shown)


.github/workflows/tiledb-go.yml
```diff
@@ -123,7 +123,6 @@ jobs:
 
   Linux_Address_Sanitizer:
     # Only run this job for releases
-    if: github.event_name == 'release' && github.event.action == 'published'
```
shaunrd0, line 126: Just testing with CI on this PR, once I see it's passing I'll revert this so it's only ran for releases.

thetorpedodog:
For [these lint errors](https://github.com/TileDB-Inc/TileDB-Go/pull/345/files#file-query_test-go-L2903), you should be able to switch them out with a call to the `slicePtr` function:

https://github.com/TileDB-Inc/TileDB-Go/blob/0b1650f2ca849f604f469a5fc8f8607352f303a1/common.go#L21-L24

array.go
```diff
@@ -389,41 +391,51 @@ func (a *Array) NonEmptyDomain() ([]NonEmptyDomain, bool, error) {
 	isDomainEmpty := true
 	nonEmptyDomains := make([]NonEmptyDomain, 0)
 	for dimIdx := uint(0); dimIdx < ndims; dimIdx++ {
-		dimension, err := domain.DimensionFromIndex(dimIdx)
-		if err != nil {
-			return nil, false, err
-		}
+		// Wrapped in a function so `dimension` will be cleaned up with defer each time the function completes.
+		err = func() error {
```
NullHypothesis, line 395: Nicely done!

NullHypothesis (approved):
Thanks for working on this! I'm surprised by how many `Free` calls were missing.

query.go
```diff
@@ -537,51 +554,61 @@ func (q *Query) EstimateBufferElements() (map[string][3]uint64, error) {
...
-			return nil, err
-		}
-
-		dimType, err := dim.Type()
-		if err != nil {
-			return nil, err
-		}
-
-		dataTypeSize := dimType.Size()
+		err = func() error {
```
thetorpedodog, line 565: here too

query_test.go
```diff
@@ -2986,8 +2984,7 @@ func TestGetDataBuffer(t *testing.T) {
 	require.NoError(t, err)
 	storedDataBuffer, ok := storedBuffer.([]int32)
 	require.True(t, ok)
-	require.Equal(t, ((*reflect.SliceHeader)(unsafe.Pointer(&dataBuffer)).Data),
-		((*reflect.SliceHeader)(unsafe.Pointer(&storedDataBuffer)).Data))
+	require.Equal(t, uintptr(slicePtr(dataBuffer)), uintptr(slicePtr(storedDataBuffer)))
```
thetorpedodog, line 2987: you should be able to remove these `uintptr` conversions as well, since both of them are now `unsafe.Pointer`s (I think?)

vfs.go
```diff
@@ -383,7 +383,7 @@ func (v *VFS) Read(fh *VFSfh, offset uint64, nbytes uint64) ([]byte, error) {
 // appends data at the end of the file. If the file does not exist,
 // it will be created.
 func (v *VFS) Write(fh *VFSfh, bytes []byte) error {
-	cbuffer := C.CBytes(bytes)
+	cbuffer := slicePtr(bytes)
```
thetorpedodog, line 386: This looks like a very, very good change: `C.CBytes` [makes a **copy** of the `[]byte` it’s given](https://pkg.go.dev/cmd/cgo#:~:text=//%20Go%20%5B%5Dbyte%20slice%20to%20C%20array%0A//%20The%20C%20array%20is%20allocated%20in%20the%20C%20heap%20using%20malloc.%0A//%20It%20is%20the%20caller%27s%20responsibility%20to%20arrange%20for%20it%20to%20be%0A//%20freed%2C%20such%20as%20by%20calling%20C.free%20(be%20sure%20to%20include%20stdlib.h%0A//%20if%20C.free%20is%20needed).%0Afunc%20C.CBytes(%5B%5Dbyte)%20unsafe.Pointer), which meant that every time we were calling `someVFS.Write(fh, ourData)`, we were _leaking a copy of our data_. Since this is now using `slicePtr`, we’re giving it the address of our existing `[]byte` and telling core to look in there.

thetorpedodog (approved):
> Thanks for working on this! I'm surprised by how many `Free` calls were missing.

I actually didn’t realize that after accessing, for example, `someArray.Schema()`, you were able/allowed to free the schema that was returned to you—I thought it belonged to the array. (Reading the code itself makes it clear that you can, but the GC can still handle it for you if you don’t care about freeing things ASAP (which arguably we should inside our own library, just for the sake of being responsible).)

vfs.go
thetorpedodog: …although I think we may also want a `defer runtime.KeepAlive(bytes)` right after this, because otherwise `bytes` is never referenced once it enters the `C.tiledb_vfs_write` call and thus would be eligible for GC as it is in progress.

query_test.go
shaunrd0: Done

query.go
shaunrd0: Done

.github/workflows/tiledb-go.yml
shaunrd0: Reverted this, but kept the removal of `continue-on-error`, there is a full CI run here with passing ASAN jobs

https://github.com/TileDB-Inc/TileDB-Go/actions/runs/10497212377/job/29079527674?pr=345

vfs.go
shaunrd0: Done, thanks!

snagles: really nice catch

PR #349, opened 2024-09-09
Remove /max_buffer_sizes support

robertbindar:
That request is being removed, more context in the REST PR here: https://github.com/TileDB-Inc/TileDB-Cloud-REST/pull/4852

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,3 +128,8 @@ The methods will be removed in the release following their removal from TileDB c
 
 All deprecated APIs in TileDB-Go are removed as the corresponding C-APIs will be removed in
 the following TileDB release.
+
+### 0.33.0
+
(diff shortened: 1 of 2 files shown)


thetorpedodog (commented):
removing the `"reflect"` import on line 13 should fix all your problems.

other than that this looks fine from a code perspective (who doesn’t love a 100% red diff?) but I would like somebody with more knowledge of the entire process to give it the final OK

NullHypothesis:
Is it worth adding a sentence to the README's section on [Deprecated Functionality](https://github.com/TileDB-Inc/TileDB-Go?tab=readme-ov-file#deprecated-functionality)?

NullHypothesis (approved):
In an ideal world, we would have first marked the function as deprecated but this should be fine.

PR #354, opened 2024-10-09
Return an error if the buffer is too large.

NullHypothesis:
So far, we would cast a uint64 to a int32 but that may result in an integer overflow. At this point, GoBytes does not seem to support buffers this large, so the next best thing to do is return an error if the given buffer is too large.

---

According to the test below, this check works correctly. I'd rather not add this (or a similar) test because it relies on allocating a 2 GiB buffer.
```go
func TestBufferOverflow(t *testing.T) {
        context, err := NewContext(nil)
        require.NoError(t, err)
        buffer, err := NewBuffer(context)
        require.NoError(t, err)

        require.NoError(t, buffer.SetBuffer(make([]byte, math.MaxInt32+10)))
        buffer.data = nil
        c, err := buffer.dataCopy()
        t.Logf("buf len: %d", len(c))
        t.Log(err)
}
```

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -8,6 +8,7 @@ import "C"
 import (
 	"bytes"
 	"fmt"
+	"math"
 	"runtime"
 	"unsafe"
 )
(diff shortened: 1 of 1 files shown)


shaunrd0 (approved):
:+1:

ypatia (approved):
Thanks for the fix! It will help uncover the places were we ask for serializing more than 2GB of data, and then we'll have to fix those.

PR #357, opened 2024-11-26
Add zero-copy serialization APIs.

teo-tsirpanis:
[SC-58009](https://app.shortcut.com/tiledb-inc/story/58009/add-apis-to-enable-zero-copy-deserialization)

This PR updates the `Buffer` and `BufferList` types to implement the [`io.WriterTo`](https://pkg.go.dev/io#WriterTo) interface, by writing the data directly to an `io.Writer` without an intermediate copy to Go-backed memory. For `Buffer` we do that with the [`Unsafe.Slice`](https://pkg.go.dev/unsafe#Slice) function (it is safe as long as the `Writer` follows the contract and does not retain the slice after the call to `Write`), and for `BufferList` we call `WriteTo` for each buffer.

In order to take advantage of this, new serialization APIs were added that return a `Buffer` instead of a `byte[]` (and unavoidably had to make a copy to Go-managed memory). The old serialization APIs, as well as the `Buffer.Serialize()` and `BufferList.Flatten()` functions were deprecated.


buffer.go
```diff
@@ -103,14 +106,61 @@ func (b *Buffer) Serialize(serializationType SerializationType) ([]byte, error)
...
+			writeSize = math.MaxInt32
+		} else {
+			writeSize = int32(remaining)
+		}
+
+		// Construct a slice from the buffer's data without copying it.
+		// Keep the buffer alive during the write, to prevent the GC from
+		// collecting the memory while it's being used.
+		n, err := w.Write(unsafe.Slice((*byte)(cbuffer), writeSize))
+		runtime.KeepAlive(b)
```
shaunrd0, line 148: I think since `b` is the object that implements this method the keepalive isn't needed

serialize.go
```diff
@@ -94,6 +106,18 @@ func SerializeArrayNonEmptyDomain(a *Array, serializationType SerializationType)
 		return nil, fmt.Errorf("Error serializing array nonempty domain: %s", a.context.LastError())
 	}
 
+	return &buffer, nil
+}
+
+// SerializeArrayNonEmptyDomainToBuffer gets and serializes the array nonempty domain and returns a Buffer object containing the payload.
```
shaunrd0, line 112:
```suggestion
// SerializeArrayNonEmptyDomain gets and serializes the array nonempty domain.
```

serialize.go
```diff
@@ -59,7 +71,7 @@ func DeserializeArraySchema(buffer *Buffer, serializationType SerializationType,
 }
 
 // SerializeArrayNonEmptyDomain gets and serializes the array nonempty domain.
```
shaunrd0, line 73:
```suggestion
// SerializeArrayNonEmptyDomainToBuffer gets and serializes the array nonempty domain and returns a Buffer object containing the payload.
```

serialize_test.go
```diff
@@ -1,11 +1,60 @@
...
+		t.Fatal(err)
+	}
+	bytes := &bytes.Buffer{}
+	_, err = buffer.WriteTo(bytes)
+	if err != nil {
+		t.Fatal(err)
+	}
+	assert.NotEmpty(t, bytes.Bytes())
+	runtime.GC()
+	runtime.GC()
```
shaunrd0, line 55: I don't understand the purpose of the last few lines in this test, but I see there was one like it just below so I'm probably missing something. 

Should there be a check here that `buffer` was garbage collected, or is the test verifying something else with the explicit calls to `runtime.GC()`?

buffer_list.go
```diff
@@ -44,6 +45,36 @@ func (b *BufferList) Context() *Context {
 	return b.context
 }
 
+// WriteTo writes the contents of a BufferList to an io.Writer.
+func (b *BufferList) WriteTo(w io.Writer) (n int64, err error) {
```
shaunrd0, line 49: Were the named return values intentional? Feel free to disregard, just pointing it out since I don't see these names being directly used in the function body.

shaunrd0 (approved):
Left some small comments but LGTM, nice work!

serialize_test.go
teo-tsirpanis: I copied the GC calls from the existing test, I don't know what their purpose was.

buffer_list.go
teo-tsirpanis: Must have been from a previous iteration of the code. Removed.

buffer.go
teo-tsirpanis: Removed. I was being extra cautious, but if we call `b.WriteTo(w)`, it becomes clear that `b` is being kept somewhere reachable.

serialize.go
teo-tsirpanis: Fixed, thanks for catching it.

ypatia (commented):
Thanks for this very useful optimization, great job! One question, is this Core [PR](https://github.com/TileDB-Inc/TileDB/pull/5353) a prerequisite? If yes let's mark this one as Draft so that we don't accidentally merge it before Go has bundled to a Core version that includes that change.

teo-tsirpanis:
Strictly speaking no, this PR can be merged without waiting for the Core PR. The Core PR is a prerequisite for fully taking advantage of the APIs added in this PR in the REST server (and even this might not be true, I have thought of a potential workaround).

buffer_test.go
```diff
@@ -95,3 +96,60 @@ func TestBufferSafety(t *testing.T) {
...
+	return len(x), nil
+}
+
+func TestWriteTo(t *testing.T) {
+	context, err := NewContext(nil)
+	require.NoError(t, err)
+	buffer, err := NewBuffer(context)
+	require.NoError(t, err)
+
+	testSizes := [5]int{0, 16, 256, 65536, 268435456}
```
NullHypothesis, line 115: Nit: I like to use separators to make large numbers easier to parse visually, but that's just personal taste. Feel free to ignore.
```suggestion
	testSizes := [5]int{0, 16, 256, 65536, 268_435_456}
```

buffer_test.go
```diff
@@ -95,3 +96,60 @@ func TestBufferSafety(t *testing.T) {
 	t.Log("post gc 2")
 	verify()
 }
+
+type ByteCounter struct {
```
NullHypothesis, line 100: Nit: I don't think we need to export this type, so probably best to have it start lower case.
```suggestion
type byteCounter struct {
```

serialize.go
```diff
@@ -464,6 +625,50 @@ func DeserializeQueryAndArray(context *Context, buffer *Buffer, serializationTyp
...
+		return errors.New("failed to retrieve bytes from buffer")
+	}
+	// cstrings are null terminated. Go's are not, add it as a suffix
+	if err := buffer.SetBuffer(append(b, []byte("\u0000")...)); err != nil {
+		return errors.New("failed to add null terminator to buffer")
+	}
+
+	ret := C.tiledb_deserialize_group_metadata(g.context.tiledbContext, g.group, C.tiledb_serialization_type_t(serializationType), buffer.tiledbBuffer)
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error deserializing group metadata: %s", g.context.LastError())
```
NullHypothesis, line 666: Nit: Error strings in Go should start with lower case because they're often chained together. I know you just copied the code, but may as well polish it up a bit!
```suggestion
		return fmt.Errorf("error deserializing group metadata: %s", g.context.LastError())
```

NullHypothesis (commented):
Neat work, thanks for this! Left a few comments.

buffer.go
```diff
@@ -103,14 +106,60 @@ func (b *Buffer) Serialize(serializationType SerializationType) ([]byte, error)
...
+	if cbuffer == nil || csize == 0 {
+		return 0, nil
+	}
+
+	remaining := int64(csize)
+
+	// Because io.Writer supports writing up to 2GB of data at a time, we have to use a loop
+	// for the bigger buffers.
+	for remaining > 0 {
+		// TODO: Use min on Go 1.21+
```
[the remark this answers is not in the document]
teo-tsirpanis: I will bump when I do SC-58351 which needs 1.21+.

buffer.go
```diff
@@ -103,14 +106,101 @@ func (b *Buffer) Serialize(serializationType SerializationType) ([]byte, error)
...
+		var writeSize int
+		if remaining > math.MaxInt {
+			writeSize = math.MaxInt
```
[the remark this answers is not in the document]
teo-tsirpanis: I used `int` because that's the type of the length of a slice. If it is 64-bits long this code should still work.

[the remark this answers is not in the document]
teo-tsirpanis: I used `int` because that's the type of the length of a slice. If it is 64-bits long this code should still work.

buffer.go
```diff
@@ -103,14 +106,101 @@ func (b *Buffer) Serialize(serializationType SerializationType) ([]byte, error)
...
+	}
+
+	if cbuffer == nil || csize == 0 {
+		return 0, nil
+	}
+
+	availableBytes := uint64(csize) - uint64(off)
+	var sizeToRead int
+	if availableBytes > math.MaxInt {
+		sizeToRead = math.MaxInt
```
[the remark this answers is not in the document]
teo-tsirpanis: I'm not sure how `io.Copy()` is relevant here as we don't have neither a `Reader` nor a `Writer` available. See also my other comment about the choice of `int` and the type of slice lengths.

buffer.go
```diff
@@ -103,14 +106,101 @@ func (b *Buffer) Serialize(serializationType SerializationType) ([]byte, error)
...
+
+	var cbuffer unsafe.Pointer
+	var csize C.uint64_t
+
+	ret := C.tiledb_buffer_get_data(b.context.tiledbContext, b.tiledbBuffer, &cbuffer, &csize)
+	if ret != C.TILEDB_OK {
+		return 0, fmt.Errorf("error getting tiledb buffer data: %w", b.context.LastError())
+	}
+
+	if uintptr(off) > uintptr(csize) {
```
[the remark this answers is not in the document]
teo-tsirpanis: Maybe, but I'm thinking whether it would be better to `return 0, io.EOF` here. [Documentation](https://pkg.go.dev/io#ReaderAt) is not clear on the behavior of `off` being out of bounds and I found [this SO question](https://stackoverflow.com/questions/76665282/behaviour-of-io-readat-at-the-end-of-input-source) which wasn't very clear either. BTW the OS file APIs (`pread` and `ReadFile`) do not fail; I ran the following C# program and it printed `0 0 0` on both Linux and Windows.

```csharp
// Creates an empty temporary file
string path = Path.GetTempFileName();

using var fh = File.OpenHandle(path, FileMode.Open, FileAccess.Read, options: FileOptions.DeleteOnClose);

byte[] buff = new byte[64];
Console.WriteLine(RandomAccess.Read(fh, buff, 0));
Console.WriteLine(RandomAccess.Read(fh, buff, 1));
Console.WriteLine(RandomAccess.Read(fh, buff, 1L << 32));
```

---

I later asked Gemini "In Go, what should the behavior of ReadAt be if the offset is out of bounds?" and it said:

> In Go, the `ReadAt` function should return an error if the offset is out of bounds. Specifically:
>
> 1. Offset is negative: It should return an error like `errors.New("offset cannot be negative")`.
>
> 2. Offset is beyond the buffer's size: It should return an error. A clear error message, like `fmt.Errorf("offset cannot be greater than buffer size")`, is helpful. If the offset equals the buffer's size, meaning there's nothing to read, it's acceptable (though not strictly required by the io.ReaderAt interface) to return 0, io.EOF to signal the end of the data.
>
> It's important _not_ to panic in these out-of-bounds situations. Return an error so the calling code can handle the situation gracefully.

But I don't know if it is biased on the code I already wrote. 😅

buffer.go
```diff
@@ -103,14 +106,101 @@ func (b *Buffer) Serialize(serializationType SerializationType) ([]byte, error)
...
+		return 0, fmt.Errorf("offset cannot be greater than buffer size")
+	}
+
+	if cbuffer == nil || csize == 0 {
```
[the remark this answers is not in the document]
teo-tsirpanis: I think it's better to make all error checking first. If the buffer is empty and we read at an invalid offset, this is still wrong (or not, see my other comment).

BTW we should return `io.EOF` as error in this case, I will change it.

buffer_list.go
```diff
@@ -44,6 +45,36 @@ func (b *BufferList) Context() *Context {
...
+// WriteTo writes the contents of a BufferList to an io.Writer.
+func (b *BufferList) WriteTo(w io.Writer) (int64, error) {
+	nbuffs, err := b.NumBuffers()
+	if err != nil {
+		return 0, err
+	}
+
+	written := int64(0)
+
+	for i := uint(0); i < uint(nbuffs); i++ {
```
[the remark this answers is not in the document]
teo-tsirpanis: `GetBuffer()` unfortunately accepts a `uint`.

buffer.go
```diff
@@ -103,14 +106,101 @@ func (b *Buffer) Serialize(serializationType SerializationType) ([]byte, error)
...
 	default:
 		return nil, fmt.Errorf("unsupported serialization type: %v", serializationType)
 	}
 	return bs, nil
 }
 
+// ReadAt writes the contents of a Buffer at a given offset to a slice.
+func (b *Buffer) ReadAt(p []byte, off int64) (int, error) {
+	if off < 0 {
+		return 0, fmt.Errorf("offset cannot be negative")
```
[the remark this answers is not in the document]
teo-tsirpanis: Done.

buffer.go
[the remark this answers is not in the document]
teo-tsirpanis: `os.File.ReadAt()` returns `0, io.EOF` in case of offset overflows and fails in case of negative offsets, as evidenced by the following test:

```go
func TestReadAtFile(t *testing.T) {
	f, err := os.CreateTemp("/tmp", "*")
	require.NoError(t, err)
	defer os.Remove(f.Name())

	buff := make([]byte, 1024)
	n, err := f.ReadAt(buff, 0)
	require.Equal(t, io.EOF, err)
	n, err = f.ReadAt(buff, 1<<32)
	require.Equal(t, io.EOF, err)
	n, err = f.ReadAt(buff, -1)
	require.NotEqual(t, io.EOF, err)

	require.Equal(t, 0, n)
}
```

I updated `Buffer.ReadAt()` to have a similar behavior.

buffer.go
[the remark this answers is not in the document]
teo-tsirpanis: Merged with the previous block.

buffer_test.go
teo-tsirpanis: Updated to `256 << 20` to more clearly state where this number comes from.

buffer_test.go
teo-tsirpanis: Done.

serialize.go
teo-tsirpanis: Updated all instances in `serialize.go`.

teo-tsirpanis:
Latest changes validated in the REST server (see CI of PR 5103, failures look unrelated).

ypatia (approved):

ypatia:
We have tested this change in Core REST-CI and it is [green](https://app.shortcut.com/tiledb-inc/story/60405/test-and-validate-pr-5103-for-avoiding-copy-in-rest-serialization-code#activity-60558). We have also successfully run SOMA Benchmarks using this PR and https://github.com/TileDB-Inc/TileDB-Cloud-REST/pull/5103 and have [demonstrated](https://app.shortcut.com/tiledb-inc/story/60405/test-and-validate-pr-5103-for-avoiding-copy-in-rest-serialization-code#activity-60639) the perf improvements.
I think it is now that testing has been completed and comments have been addressed, it is safe to get merged.

PR #359, opened 2024-12-13
Wrap TileDB 2.27.0 APIs

shaunrd0:
This wraps TileDB core APIs added in 2.27.0

### DIFF
--- a/enumeration_experimental.go
+++ b/enumeration_experimental.go
@@ -393,20 +393,20 @@ func (a *ArraySchema) EnumerationFromName(name string) (*Enumeration, error) {
 	defer C.free(unsafe.Pointer(cName))
 	ret := C.tiledb_array_schema_get_enumeration_from_name(a.context.tiledbContext, a.tiledbArraySchema, cName, &enum.tiledbEnum)
 	if ret != C.TILEDB_OK {
-		return nil, fmt.Errorf("Error getting enumeration from name: %s", a.context.LastError())
+		return nil, fmt.Errorf("error getting enumeration from name: %w", a.context.LastError())
 	}
 	freeOnGC(enum)
 	return enum, nil
 }
 
-// EnumerationFromName gets an Enumeration from the ArraySchema by it's Attribute name
+// EnumerationFromName gets an Enumeration from the ArraySchema by its Attribute name.
 func (a *ArraySchema) EnumerationFromAttributeName(name string) (*Enumeration, error) {
 	enum := &Enumeration{context: a.context}
 	cName := C.CString(name)
 	defer C.free(unsafe.Pointer(cName))
 	ret := C.tiledb_array_schema_get_enumeration_from_attribute_name(a.context.tiledbContext, a.tiledbArraySchema, cName, &enum.tiledbEnum)
 	if ret != C.TILEDB_OK {
-		return nil, fmt.Errorf("Error getting enumeration from attribute name: %s", a.context.LastError())
+		return nil, fmt.Errorf("error getting enumeration from attribute name: %w", a.context.LastError())
 	}
 	freeOnGC(enum)
 	return enum, nil
@@ -427,7 +427,7 @@ func (a *Array) LoadAllEnumerations() error {
 func (a *Array) LoadEnumerationsAllSchemas() error {
 	ret := C.tiledb_array_load_enumerations_all_schemas(a.context.tiledbContext, a.tiledbArray)
 	if ret != C.TILEDB_OK {
-		return fmt.Errorf("Error loading enumerations for all schemas: %s", a.context.LastError())
+		return fmt.Errorf("error loading enumerations for all schemas: %w", a.context.LastError())
 	}
 
 	return nil
--- a/enumeration_experimental.go
+++ b/enumeration_experimental.go
@@ -385,12 +385,46 @@ func (a *ArraySchema) AddEnumeration(e *Enumeration) error {
 	return nil
 }
 
+func (a *ArraySchema) EnumerationFromName(name string) (*Enumeration, error) {
+	enum := &Enumeration{context: a.context}
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+	ret := C.tiledb_array_schema_get_enumeration_from_name(a.context.tiledbContext, a.tiledbArraySchema, cName, &enum.tiledbEnum)
+	if ret != C.TILEDB_OK {
+		return nil, fmt.Errorf("Error getting enumeration from name: %s", a.context.LastError())
+	}
+	freeOnGC(enum)
+	return enum, nil
+}
+
+func (a *ArraySchema) EnumerationFromAttributeName(name string) (*Enumeration, error) {
+	enum := &Enumeration{context: a.context}
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+	ret := C.tiledb_array_schema_get_enumeration_from_attribute_name(a.context.tiledbContext, a.tiledbArraySchema, cName, &enum.tiledbEnum)
+	if ret != C.TILEDB_OK {
+		return nil, fmt.Errorf("Error getting enumeration from attribute name: %s", a.context.LastError())
+	}
+	freeOnGC(enum)
+	return enum, nil
+}
+
 // LoadAllEnumeration is for use with TileDB cloud arrays. It fetches the enumeration values from the server.
 // The method is called ondemand if the client tries to fetch enumeration values for a tiledb:// array.
 func (a *Array) LoadAllEnumerations() error {
 	ret := C.tiledb_array_load_all_enumerations(a.context.tiledbContext, a.tiledbArray)
 	if ret != C.TILEDB_OK {
 		return fmt.Errorf("Error loading all enumerations: %s", a.context.LastError())
+	}
+
+	return nil
+}
+
+// LoadEnumerationsAllSchemas is for use with TileDB cloud arrays. It fetches the enumeration values from the server for all array schemas, past and present.
+func (a *Array) LoadEnumerationsAllSchemas() error {
+	ret := C.tiledb_array_load_enumerations_all_schemas(a.context.tiledbContext, a.tiledbArray)
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error loading enumerations for all schemas: %s", a.context.LastError())
 	}
 
 	return nil
--- a/enumeration_experimental.go
+++ b/enumeration_experimental.go
@@ -155,6 +155,7 @@ func (e *Enumeration) Name() (string, error) {
 	if ret != C.TILEDB_OK {
 		return "", fmt.Errorf("Error getting name: %s", e.context.LastError())
 	}
+	defer C.tiledb_string_free(&str)
 
 	var cName *C.char
 	var cNameSize C.size_t
@@ -468,6 +469,7 @@ func (a *Attribute) GetEnumerationName() (string, error) {
 	if ret != C.TILEDB_OK {
 		return "", fmt.Errorf("Error getting enumeration name: %s", a.context.LastError())
 	}
+	defer C.tiledb_string_free(&str)
 
 	var cName *C.char
 	var cNameSize C.size_t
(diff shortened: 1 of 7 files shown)


group.go
```diff
@@ -436,3 +449,36 @@ func (g *Group) Delete(recursive bool) error {
...
+func (g *Group) IsOpen() (bool, error) {
+	var isOpen C.int32_t
+
+	ret := C.tiledb_group_is_open(g.context.tiledbContext, g.group, &isOpen)
+	if ret != C.TILEDB_OK {
+		return false, fmt.Errorf("Error checking if group is open: %s", g.context.LastError())
+	}
+
+	return isOpen > 0, nil
+}
```
teo-tsirpanis, line 484: I would either return false or panic here in case of an error; I think it's excessive to return `(T, error)` for every single API regardless of how [trivial](https://github.com/TileDB-Inc/TileDB/blob/c0458631faeeeb31f83ff4ec1ef8b3f94606c6c2/tiledb/api/c_api/group/group_api.cc#L390-L398) it is, and has made using the API difficult. Several other similar APIs come to my mind and we should consider changing them in the future.

group.go
```diff
@@ -388,21 +388,34 @@ func (g *Group) GetMetadataFromIndexWithValueLimit(index uint64, limit *uint) (*
...
 	var cRecurse C.uint8_t
 	if recurse {
 		cRecurse = 1
 	}
 
-	ret := C.tiledb_group_dump_str(g.context.tiledbContext, g.group, &cOutput, cRecurse)
+	ret := C.tiledb_group_dump_str_v2(g.context.tiledbContext, g.group, &tdbString, cRecurse)
 	if ret != C.TILEDB_OK {
 		return "", fmt.Errorf("Error dumping group contents: %s", g.context.LastError())
 	}
```
teo-tsirpanis, line 411:
```suggestion
	}
	defer C.tiledb_string_free(&tdbString)
```

enumeration_experimental.go
```diff
@@ -385,6 +385,32 @@ func (a *ArraySchema) AddEnumeration(e *Enumeration) error {
...
+	enum := &Enumeration{context: a.context}
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+	ret := C.tiledb_array_schema_get_enumeration_from_attribute_name(a.context.tiledbContext, a.tiledbArraySchema, cName, &enum.tiledbEnum)
+	if ret != C.TILEDB_OK {
+		return nil, fmt.Errorf("Error getting enumeration from attribute name: %s", a.context.LastError())
+	}
+	freeOnGC(enum)
+	return enum, nil
+}
```
teo-tsirpanis, line 412: I completely missed these when they were added. 😅

fragment_info.go
```diff
@@ -570,6 +570,17 @@ func (fI *FragmentInfo) DumpSTDOUT() error {
 	return nil
 }
 
+// String retrieves the string representation of the FragmentInfo
+func (fI *FragmentInfo) String() (string, error) {
```
teo-tsirpanis, line 574: Same with returning `error` here. Returning just a string would enable `FragmentInfo` to participate in [string formatting](https://pkg.go.dev/fmt). We shouldn't panic in this case, but return the error string or an `<ERROR>` placeholder.

shaunrd0: Done, thanks for the review!

group.go
shaunrd0: Done

group.go
shaunrd0: Nice catch, I found some more of these and freed those as well

teo-tsirpanis (approved):
Thanks! I would like a second opinion about not returning error on `Group.IsOpen()`.

enumeration_experimental.go
```diff
@@ -385,6 +386,32 @@ func (a *ArraySchema) AddEnumeration(e *Enumeration) error {
...
 }
 
+// EnumerationFromName gets an Enumeration from the ArraySchema by name
+func (a *ArraySchema) EnumerationFromName(name string) (*Enumeration, error) {
+	enum := &Enumeration{context: a.context}
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+	ret := C.tiledb_array_schema_get_enumeration_from_name(a.context.tiledbContext, a.tiledbArraySchema, cName, &enum.tiledbEnum)
+	if ret != C.TILEDB_OK {
+		return nil, fmt.Errorf("Error getting enumeration from name: %s", a.context.LastError())
```
NullHypothesis, line 396: Nit: Errors should be lower case (see [Go's documentation](https://go.dev/wiki/CodeReviewComments#error-strings)) and we should use `%w` for error wrapping, which allows the caller to use `errors.{Unwrap,Is,As}`.
```suggestion
		return nil, fmt.Errorf("error getting enumeration from name: %w", a.context.LastError())
```

enumeration_experimental.go
```diff
@@ -385,6 +386,32 @@ func (a *ArraySchema) AddEnumeration(e *Enumeration) error {
...
+}
+
+// EnumerationFromName gets an Enumeration from the ArraySchema by it's Attribute name
+func (a *ArraySchema) EnumerationFromAttributeName(name string) (*Enumeration, error) {
+	enum := &Enumeration{context: a.context}
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+	ret := C.tiledb_array_schema_get_enumeration_from_attribute_name(a.context.tiledbContext, a.tiledbArraySchema, cName, &enum.tiledbEnum)
+	if ret != C.TILEDB_OK {
+		return nil, fmt.Errorf("Error getting enumeration from attribute name: %s", a.context.LastError())
```
NullHypothesis, line 409: Same as above.
```suggestion
		return nil, fmt.Errorf("error getting enumeration from attribute name: %w", a.context.LastError())
```

enumeration_experimental.go
```diff
@@ -396,6 +423,16 @@ func (a *Array) LoadAllEnumerations() error {
 	return nil
 }
 
+// LoadEnumerationsAllSchemas is for use with TileDB cloud arrays. It fetches the enumeration values from the server for all array schemas, past and present.
+func (a *Array) LoadEnumerationsAllSchemas() error {
+	ret := C.tiledb_array_load_enumerations_all_schemas(a.context.tiledbContext, a.tiledbArray)
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error loading enumerations for all schemas: %s", a.context.LastError())
```
NullHypothesis, line 430: Same as above.
```suggestion
		return fmt.Errorf("error loading enumerations for all schemas: %w", a.context.LastError())
```

fragment_info.go
```diff
@@ -570,6 +570,23 @@ func (fI *FragmentInfo) DumpSTDOUT() error {
 	return nil
 }
 
+// String retrieves the string representation of the FragmentInfo
+func (fI *FragmentInfo) String() string {
+	var tdbString *C.tiledb_string_t
+
+	ret := C.tiledb_fragment_info_dump_str(fI.context.tiledbContext, fI.tiledbFragmentInfo, &tdbString)
+	if ret != C.TILEDB_OK {
+		return fmt.Sprintf("Error dumping fragment info to string: %s", fI.context.LastError())
```
NullHypothesis, line 579: Shouldn't we return an error here instead? Doesn't seem right that we only return a string but the string can contain an error.

group.go
```diff
@@ -388,21 +388,31 @@ func (g *Group) GetMetadataFromIndexWithValueLimit(index uint64, limit *uint) (*
 	return &groupMetadata, nil
 }
 
+// Dump the Group to a string value
 func (g *Group) Dump(recurse bool) (string, error) {
-	var cOutput *C.char
-	defer C.free(unsafe.Pointer(cOutput))
+	if !g.IsOpen() {
+		return "", fmt.Errorf("Error dumping group to string: Group must be opened in TILEDB_READ mode")
```
NullHypothesis, line 394:
```suggestion
		return "", fmt.Errorf("error dumping group to string: group must be opened in TILEDB_READ mode")
```

group.go
```diff
@@ -388,21 +388,31 @@ func (g *Group) GetMetadataFromIndexWithValueLimit(index uint64, limit *uint) (*
...
+	ret := C.tiledb_group_dump_str_v2(g.context.tiledbContext, g.group, &tdbString, cRecurse)
 	if ret != C.TILEDB_OK {
 		return "", fmt.Errorf("Error dumping group contents: %s", g.context.LastError())
 	}
+	defer C.tiledb_string_free(&tdbString)
 
-	return C.GoString(cOutput), nil
+	dumpStr, err := stringHandleToString(tdbString)
+	if err != nil {
+		return "", fmt.Errorf("Error dumping group contents: %s", g.context.LastError())
```
NullHypothesis, line 412:
```suggestion
		return "", fmt.Errorf("error dumping group contents: %w", g.context.LastError())
```

group.go
```diff
@@ -436,3 +446,36 @@ func (g *Group) Delete(recursive bool) error {
 	}
 	return nil
 }
+
+// AddMemberWithType adds a member to the Group providing its type
```
NullHypothesis, line 450: Was this comment cut off? 🤔 Either way, I would add a period.

group.go
```diff
@@ -436,3 +446,36 @@ func (g *Group) Delete(recursive bool) error {
...
+// This method is recommended for performance when operating on remote groups
```
NullHypothesis, line 451:
```suggestion
// This method is recommended for performance when operating on remote groups.
```

group.go
```diff
@@ -436,3 +446,36 @@ func (g *Group) Delete(recursive bool) error {
...
+	defer C.free(unsafe.Pointer(cname))
+
+	var cRelative C.uint8_t
+	if isRelativeURI {
+		cRelative = 1
+	}
+
+	ret := C.tiledb_group_add_member_with_type(g.context.tiledbContext, g.group, curi, cRelative, cname, C.tiledb_object_t(objectType))
+	if ret != C.TILEDB_OK {
+		return fmt.Errorf("Error adding member with type to group: %s", g.context.LastError())
```
NullHypothesis, line 466:
```suggestion
		return fmt.Errorf("error adding member with type to group: %w", g.context.LastError())
```

group.go
```diff
@@ -436,3 +446,36 @@ func (g *Group) Delete(recursive bool) error {
...
+	}
+	return nil
+}
+
+// IsOpen returns true if the Group is open or false if the group is closed
```
NullHypothesis, line 471:
```suggestion
// IsOpen returns true if the Group is open or false if the group is closed.
```

group.go
```diff
@@ -436,3 +446,36 @@ func (g *Group) Delete(recursive bool) error {
...
+func (g *Group) IsOpen() bool {
+	var isOpen C.int32_t
+
+	ret := C.tiledb_group_is_open(g.context.tiledbContext, g.group, &isOpen)
+	if ret != C.TILEDB_OK {
+		panic(fmt.Errorf("Error checking if group is open: %s", g.context.LastError()))
```
NullHypothesis, line 477: I'm not a big fan of panicking here. We tend to only use panics if recovery is impossible. For example, REST panics if it cannot run database migrations upon starting. This doesn't strike me as a case where recovery is impossible, so I'd prefer to bubble up an error here.

query_experimental.go
```diff
@@ -65,6 +65,7 @@ func (q *Query) GetPlan() (string, error) {
 	if ret != C.TILEDB_OK {
 		return "", fmt.Errorf("Error getting query plan: %s", q.context.LastError())
 	}
+	defer C.tiledb_string_free(&plan)
```
NullHypothesis, line 68: ❤️

enumeration_experimental.go
```diff
@@ -432,6 +469,7 @@ func (a *Attribute) GetEnumerationName() (string, error) {
 	if ret != C.TILEDB_OK {
 		return "", fmt.Errorf("Error getting enumeration name: %s", a.context.LastError())
 	}
+	defer C.tiledb_string_free(&str)
```
NullHypothesis, line 472: 😍

NullHypothesis (commented):
Thanks for this! Overall looking great. Left a bunch of hopefully useful comments.

fragment_info.go
shaunrd0: Done, thanks for taking a look

group.go
shaunrd0: Done

NullHypothesis (approved):
Thanks for the changes! A work of art 👌

PR #363, opened 2024-12-20
Add wrapper for `tiledb_vfs_ls_recursive`.

teo-tsirpanis:
[SC-61002](https://app.shortcut.com/tiledb-inc/story/61002/support-tiledb-vfs-ls-recursive-in-go)

### DIFF
--- a/clibrary.h
+++ b/clibrary.h
@@ -10,31 +10,31 @@ int32_t vfsLs(cchar_t* path, void *data);
 int32_t vfsLsRecursive(cchar_t* path, size_t path_len, uint64_t size, void *data);
 int32_t objectsInPath(cchar_t* path, tiledb_object_t objectType, void *data);
 
-TILEDB_EXPORT int32_t _num_of_folders_in_path(
+int32_t _num_of_folders_in_path(
     tiledb_ctx_t* ctx,
     tiledb_vfs_t* vfs,
     const char* path,
     void* data);
 
-TILEDB_EXPORT int32_t _vfs_ls(
+int32_t _vfs_ls(
     tiledb_ctx_t* ctx,
     tiledb_vfs_t* vfs,
     const char* path,
     void* data);
 
-TILEDB_EXPORT int32_t _vfs_ls_recursive(
+int32_t _vfs_ls_recursive(
     tiledb_ctx_t* ctx,
     tiledb_vfs_t* vfs,
     const char* path,
     void* data);
 
-TILEDB_EXPORT int32_t _tiledb_object_walk(
+int32_t _tiledb_object_walk(
     tiledb_ctx_t* ctx,
     const char* path,
     tiledb_walk_order_t order,
     void* data);
 
-TILEDB_EXPORT int32_t _tiledb_object_ls(
+int32_t _tiledb_object_ls(
     tiledb_ctx_t* ctx,
     const char* path,
     void* data);
--- a/.github/workflows/tiledb-go.yml
+++ b/.github/workflows/tiledb-go.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         # Will be checking following versions
-        go: ["1.20", "1.21", "1.22", "1.23"]
+        go: ["1.22", "1.23"]
     steps:
 
     # Checks out repository
@@ -72,7 +72,7 @@ jobs:
     strategy:
       matrix:
         # Will be checking following versions
-        go: ["1.20", "1.21", "1.22", "1.23"]
+        go: ["1.22", "1.23"]
     steps:
     # Checks out repository
     - uses: actions/checkout@v4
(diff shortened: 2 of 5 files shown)


vfs.go
```diff
@@ -464,7 +464,7 @@ func (v *VFS) NumOfFragmentsInPath(path string) (int, error) {
 		Vfs:          v,
 	}
 	data := pointer.Save(&numOfFragmentsData)
-	defer C.free(data)
+	defer pointer.Unref(data)
```
NullHypothesis, line 467: ❤️

vfs.go
```diff
@@ -635,6 +635,7 @@ func (v *VFS) List(path string) ([]string, []string, error) {
 		Vfs:     v,
 	}
 	data := pointer.Save(&folderData)
+	defer pointer.Unref(data)
```
NullHypothesis, line 638: 😍

vfs.go
```diff
@@ -643,3 +644,52 @@ func (v *VFS) List(path string) ([]string, []string, error) {
 
 	return folderData.Folders, folderData.Files, nil
 }
+
+type visitRecursiveCallback = func(path string, size uint64) (doContinue bool, err error)
```
NullHypothesis, line 648: We might want to export this type because `VisitRecursive` may otherwise be annoying to use: Not sure if all IDEs display non-exported types. If an IDE doesn't, you have to look at TileDB-Go's source code to figure out what function signature is expected here.
```suggestion
type VisitRecursiveCallback = func(path string, size uint64) (doContinue bool, err error)
```

NullHypothesis (commented):
Left a minor comment but looks good otherwise. I'll defer to Shaun or Ypatia for a sign-off considering that they have actual core expertise.

vfs.go
teo-tsirpanis: I didn't want to copy-paste the callback's signature, but also didn't want to export a type without it being needed, but if some IDEs might not like it sure, done.

vfs.go
```diff
@@ -643,3 +644,52 @@ func (v *VFS) List(path string) ([]string, []string, error) {
...
+//export vfsLsRecursive
+func vfsLsRecursive(path *C.cchar_t, path_len C.size_t, size C.uint64_t, data unsafe.Pointer) int32 {
+	state := pointer.Restore(data).(*visitRecursiveState)
+
+	if path_len > math.MaxInt {
+		state.lastError = errors.New("path is too long")
+		return 0
+	}
+
+	doContinue, err := state.callback(C.GoStringN(path, C.int(path_len)), uint64(size))
```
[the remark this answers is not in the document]
shaunrd0: Do you mean using `recover()` to catch and rethrow an `error` if the callback panics? IMO if a user writes a callback that panics that's what should happen but @NullHypothesis please correct me if I am wrong.

For example I tried this in `examples/vfs.go` and it panics, I think that's what would be expected.

```golang
err = vfs.VisitRecursive(dirA, func(path string, size uint64) (bool, error) {
    panic("Test")
    return false, nil
})
```

clibrary.h
```diff
@@ -1,12 +1,13 @@
 #ifndef CLIBRARY_H
 #define CLIBRARY_H
 
-#include <tiledb/tiledb.h>
+#include <tiledb/tiledb_experimental.h>
```
shaunrd0, line 4: We could create a `vfs_experimental.go` and `clibrary_experimental.{h,c}` to keep non experimental and experimental features separate

shaunrd0 (commented):
Just one comment for experimental things, LGTM otherwise :+1:

vfs.go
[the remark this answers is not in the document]
teo-tsirpanis: Yes it should still panic, but I'm not sure what is the behavior if we panic in a Go function that is called from C code, and whether we should have additional logic to handle it. I know that in C# this is not supported (because there is no standardized ABI for exceptions in non-Windows), but don't know anything about that in Go.

clibrary.h
teo-tsirpanis: We could, but I'm not sure what the benefit would be. We can add a `//go:build experimental` at the top of the file like [this file](https://github.com/TileDB-Inc/TileDB-Go/pull/225/files#diff-0d15a00563eb8bca6ddb7ef55eb18247159f8419c17be402ab23d65330d319d9) in #225 does, but currently we are not doing this in the existing `experimental` files.

shaunrd0: Yeah, it looks like the experimental tag was removed from REST here https://github.com/TileDB-Inc/TileDB-Cloud-REST/pull/4472 and then we removed the `experimental` build tag and added the `_experimental.go` suffix to TileDB-Go sources in https://github.com/TileDB-Inc/TileDB-Go/pull/302

Initially my vote was to add the build flags that were missing at the top of these `_experimental.go` files, but after seeing the above PRs it looks like we actually don't want the experimental tag and it should not be present at the top of these files, otherwise if we added it REST will be missing features it depends on. So.. LGTM! :+1:

shaunrd0 (approved):

PR #366, opened 2025-02-14
Improve memory safety of cgo interop code.

teo-tsirpanis:
[SC-58351](https://app.shortcut.com/tiledb-inc/story/58351)
[SC-62492](https://app.shortcut.com/tiledb-inc/story/62492)

* We use `runtime.KeepAlive` to keep the finalizers of TileDB objects from being called during each cgo call.
  * The keepalives are put right after the call, ordered by the object's appearence in the C function's parameters.
  * If a C API returns pointers to C-owned memory, these are documented, and the keepalives are put right after the last use of that pointer.
  * If a variable does not need a keepalive when using it or passing it to C, that fact is documented.
* We use the `runtime.Pinner` type to pin memory held long-term by C, in a well-defined way.
  * Unpinning happens in the `Free()` function, after the C handle is freed.
  * This makes us (to the best of my knowledge) no longer reliant on a non-moving garbage collector. I'm not removing the dependency to https://go4.org/unsafe/assume-no-moving-gc right now out of abundance of caution.
* Updates CI to build tests with the `GOEXPERIMENT=cgocheck2` option.

### DIFF
--- a/.github/workflows/tiledb-go.yml
+++ b/.github/workflows/tiledb-go.yml
@@ -65,6 +65,8 @@ jobs:
 
     # Tests TileDB-Go
     - name: Test TileDB-Go
+      env:
+        GOEXPERIMENT: cgocheck2
       run: go test -gcflags=all=-d=checkptr=2 -v ./...
 
   Macos_Test:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,3 +1,3 @@
-run:
-  skip-dirs:
+issues:
+  exclude-dirs:
     - cmd
--- a/array.go
+++ b/array.go
@@ -118,6 +118,7 @@ func WithStartTime(start time.Time) ArrayOpenOption {
 func WithEndTimestamp(endTimestamp uint64) ArrayOpenOption {
 	return func(tdbArray *Array) error {
 		ret := C.tiledb_array_set_open_timestamp_end(tdbArray.context.tiledbContext, tdbArray.tiledbArray, C.uint64_t(endTimestamp))
+		runtime.KeepAlive(tdbArray)
 		if ret != C.TILEDB_OK {
 			return fmt.Errorf("error setting end timestamp option: %w", tdbArray.context.LastError())
 		}
@@ -129,6 +130,7 @@ func WithEndTimestamp(endTimestamp uint64) ArrayOpenOption {
 func WithStartTimestamp(startTimestamp uint64) ArrayOpenOption {
 	return func(tdbArray *Array) error {
 		ret := C.tiledb_array_set_open_timestamp_start(tdbArray.context.tiledbContext, tdbArray.tiledbArray, C.uint64_t(startTimestamp))
+		runtime.KeepAlive(tdbArray)
 		if ret != C.TILEDB_OK {
 			return fmt.Errorf("error setting start timestamp option: %w", tdbArray.context.LastError())
 		}
@@ -154,6 +156,7 @@ func (a *Array) OpenWithOptions(queryType QueryType, opts ...ArrayOpenOption) er
 	}
 
 	ret := C.tiledb_array_open(a.context.tiledbContext, a.tiledbArray, C.tiledb_query_type_t(queryType))
+	runtime.KeepAlive(a)
 	if ret != C.TILEDB_OK {
 		return fmt.Errorf("error opening tiledb array for querying: %w", a.context.LastError())
 	}
@@ -172,6 +175,7 @@ creation and submission of queries for both these array objects.
 */
 func (a *Array) Open(queryType QueryType) error {
 	ret := C.tiledb_array_open(a.context.tiledbContext, a.tiledbArray, C.tiledb_query_type_t(queryType))
+	runtime.KeepAlive(a)
 	if ret != C.TILEDB_OK {
 		return fmt.Errorf("error opening tiledb array for querying: %w", a.context.LastError())
 	}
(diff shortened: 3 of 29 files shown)


subarray.go
```diff
@@ -182,16 +187,14 @@ func (sa *Subarray) AddRange(dimIdx uint32, r Range) error {
...
 	} else {
 		startValue := addressableValue(r.start)
 		endValue := addressableValue(r.end)
 		ret = C.tiledb_subarray_add_range(sa.context.tiledbContext, sa.subarray, C.uint32_t(dimIdx),
 			startValue.UnsafePointer(), endValue.UnsafePointer(), nil)
-		runtime.KeepAlive(startValue)
-		runtime.KeepAlive(endValue)
 	}
+	runtime.KeepAlive(sa)
+	// The start and end pointers are being kept alive by passing them to cgo calls.
```
teo-tsirpanis, line 197: From [documentation](https://pkg.go.dev/cmd/cgo#hdr-Passing_pointers):

> Go pointers passed as function arguments to C functions have the memory they point to implicitly pinned for the duration of the call.

buffer.go
```diff
@@ -211,28 +200,12 @@ func (b *Buffer) dataCopy() ([]byte, error) {
...
-		defer runtime.KeepAlive(b)
-
-		if csize > math.MaxInt32 {
-			return nil, fmt.Errorf("TileDB's buffer (%d) larger than maximum allowed CGo buffer (%d)", csize, math.MaxInt32)
-		}
-		return C.GoBytes(cbuffer, C.int(csize)), nil
+	if csize > math.MaxInt32 {
+		return nil, fmt.Errorf("TileDB's buffer (%d) larger than maximum allowed CGo buffer (%d)", csize, math.MaxInt32)
 	}
+	cpy := C.GoBytes(cbuffer, C.int(csize))
```
teo-tsirpanis, line 206: We can't know whether the memory is owned by Go or C, so we always make a copy. This API is not recommended either way.

query.go
```diff
@@ -300,12 +304,10 @@ func (q *Query) SetQueryCondition(cond *QueryCondition) error {
 // for any other query type.
 func (q *Query) Finalize() error {
 	ret := C.tiledb_query_finalize(q.context.tiledbContext, q.tiledbQuery)
+	runtime.KeepAlive(q)
 	if ret != C.TILEDB_OK {
 		return fmt.Errorf("error finalizing query: %w", q.context.LastError())
 	}
-	q.bufferMutex.Lock()
-	defer q.bufferMutex.Unlock()
-	q.buffers = nil
```
teo-tsirpanis, line 308: We should not be unpinning the buffers in `Finalize`.

array.go
```diff
@@ -118,6 +118,7 @@ func WithStartTime(start time.Time) ArrayOpenOption {
 func WithEndTimestamp(endTimestamp uint64) ArrayOpenOption {
 	return func(tdbArray *Array) error {
 		ret := C.tiledb_array_set_open_timestamp_end(tdbArray.context.tiledbContext, tdbArray.tiledbArray, C.uint64_t(endTimestamp))
+		runtime.KeepAlive(tdbArray)
```
teo-tsirpanis, line 121: When in the future we support Go 1.24's [`AddCleanup`](https://pkg.go.dev/runtime@master#AddCleanup), we would have to explicitly keep the contexts alive, because the finalizers do not run in order.

Let's do it now, before marking the PR as ready for review.

serialize.go
```diff
@@ -533,6 +555,8 @@ func DeserializeFragmentInfo(fragmentInfo FragmentInfo, buffer *Buffer, arrayURI
 	defer C.free(unsafe.Pointer(cArrayURI))
 
 	ret := C.tiledb_deserialize_fragment_info(fragmentInfo.context.tiledbContext, buffer.tiledbBuffer, C.tiledb_serialization_type_t(serializationType), cArrayURI, cClientSide, fragmentInfo.tiledbFragmentInfo)
+	runtime.KeepAlive(fragmentInfo)
```
teo-tsirpanis, line 558: This `FragmentInfo` (and the one in line 581) is passed by value, so this call does not do anything. We might want to change it now or in the future to take a pointer, don't know if the break is tolerable.

array.go
teo-tsirpanis: On second thought, I have misread documentation. `runtime.KeepAlive` does apply recursively (obviously), and as long as an object is kept alive, its finalizers will not run. This is what we want, and the order finalizers execute does not matter here.

.github/workflows/tiledb-go.yml
```diff
@@ -65,6 +65,8 @@ jobs:
 
     # Tests TileDB-Go
     - name: Test TileDB-Go
+      env:
+        GOEXPERIMENT: cgocheck2
```
shaunrd0, line 69: Was this just preventative or did this produce some failures or output prior to these changes? I tried using this on `master` and the tests pass with no additional output, just curious if you had a different experience

.golangci.yml
```diff
@@ -1,3 +1,3 @@
-run:
-  skip-dirs:
+issues:
+  exclude-dirs:
```
shaunrd0, line 2: Nice, thanks for fixing this! :slightly_smiling_face:

shaunrd0 (approved):
LGTM! :+1:

.github/workflows/tiledb-go.yml
teo-tsirpanis: It's preventative, even though it might not matter in practice since we use `unsafe.Pointer` in lots of places, if I correctly understood what this switch does.

PR #367, opened 2025-03-08
Add standalone functions for certain operations on arrays and groups.

teo-tsirpanis:
[SC-64187](https://app.shortcut.com/tiledb-inc/story/64187/add-standalone-apis-to-create-arrays-and-groups)

This PR adds the following functions to TileDB-Go:

```go
func CreateArray(context *tiledb.Context, uri string, schema *tiledb.ArraySchema) error

func CreateGroup(context *tiledb.Context, uri string)

func ConsolidateArray(tdbCtx *Context, uri string, config *Config) error

func VacuumArray(tdbCtx *Context, uri string, config *Config) error
```

Before that, in order to create an array or a group, or consolidate or vacuum an array, you had to first allocate an object with the `New***` function. This is inconsistent with what the C API exposes and what other APIs do.

The changes themselves are pretty straightforward; the majority of changes in this PR are to update all tests and examples to use the new APIs.

### DIFF
--- a/array.go
+++ b/array.go
@@ -221,7 +221,7 @@ func (a *Array) Close() error {
 	return nil
 }
 
-// Create creates a new TileDB array given a context, URI and schema.
+// CreateArray creates a new TileDB array given a context, URI and schema.
 func CreateArray(tdbCtx *Context, uri string, arraySchema *ArraySchema) error {
 	curi := C.CString(uri)
 	defer C.free(unsafe.Pointer(curi))
--- a/array.go
+++ b/array.go
@@ -208,17 +208,22 @@ func (a *Array) Close() error {
 	return nil
 }
 
-// Create creates a new TileDB array given an input schema.
-func (a *Array) Create(arraySchema *ArraySchema) error {
-	curi := C.CString(a.uri)
+// Create creates a new TileDB array given a context, URI and schema.
+func CreateArray(tdbCtx *Context, uri string, arraySchema *ArraySchema) error {
+	curi := C.CString(uri)
 	defer C.free(unsafe.Pointer(curi))
-	ret := C.tiledb_array_create(a.context.tiledbContext, curi, arraySchema.tiledbArraySchema)
-	runtime.KeepAlive(a)
+	ret := C.tiledb_array_create(tdbCtx.tiledbContext, curi, arraySchema.tiledbArraySchema)
+	runtime.KeepAlive(tdbCtx)
 	runtime.KeepAlive(arraySchema)
 	if ret != C.TILEDB_OK {
-		return fmt.Errorf("error creating tiledb array: %w", a.context.LastError())
+		return fmt.Errorf("error creating tiledb array: %w", tdbCtx.LastError())
 	}
 	return nil
+}
+
+// Create creates a new TileDB array given an input schema.
+func (a *Array) Create(arraySchema *ArraySchema) error {
+	return CreateArray(a.context, a.uri, arraySchema)
 }
 
 // Consolidate consolidates the fragments of an array into a single fragment.
--- a/array.go
+++ b/array.go
@@ -69,6 +69,59 @@ func newArrayFromHandle(tdbCtx *Context, arrayHandle arrayHandle) *Array {
 	return &Array{context: tdbCtx, tiledbArray: arrayHandle}
 }
 
+// ConsolidateArray consolidates the fragments of an array into a single fragment.
+// You must first finalize all queries to the array before consolidation can
+// begin (as consolidation temporarily acquires an exclusive lock on the array).
+func ConsolidateArray(tdbCtx *Context, uri string, config *Config) error {
+	if config == nil {
+		return errors.New("Config must not be nil for Consolidate")
+	}
+
+	curi := C.CString(uri)
+	defer C.free(unsafe.Pointer(curi))
+	ret := C.tiledb_array_consolidate(tdbCtx.tiledbContext.Get(), curi, config.tiledbConfig.Get())
+	runtime.KeepAlive(tdbCtx)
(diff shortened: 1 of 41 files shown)


array.go
```diff
@@ -208,19 +208,24 @@ func (a *Array) Close() error {
...
+	runtime.KeepAlive(tdbCtx)
 	runtime.KeepAlive(arraySchema)
 	if ret != C.TILEDB_OK {
-		return fmt.Errorf("error creating tiledb array: %w", a.context.LastError())
+		return fmt.Errorf("error creating tiledb array: %w", tdbCtx.LastError())
 	}
 	return nil
 }
 
+// Create creates a new TileDB array given an input schema.
```
teo-tsirpanis, line 224:
```suggestion
// Create creates a new TileDB array given an input schema.
// Deprecated: Use CreateArray instead.
```

group.go
```diff
@@ -36,19 +36,24 @@ func NewGroup(tdbCtx *Context, uri string) (*Group, error) {
 	return &group, nil
 }
 
-// Create creates a new TileDB group.
-func (g *Group) Create() error {
-	curi := C.CString(g.uri)
+// Create creates a new TileDB group given a context and URI.
```
teo-tsirpanis, line 39:
```suggestion
// Create creates a new TileDB group given a context and URI.
// Deprecated: Use CreateGroup instead.
```

teo-tsirpanis:
There are more functions in a similar situation, like consolidating and vacuuming.

array.go
```diff
@@ -208,59 +208,76 @@ func (a *Array) Close() error {
...
 
 	runtime.KeepAlive(config)
 	return nil
 }
 
-// Vacuum cleans up the array, such as consolidated fragments and array metadata.
-func (a *Array) Vacuum(config *Config) error {
+// Consolidate consolidates the fragments of the array into a single fragment.
+// You must first finalize all queries to the array before consolidation can
+// begin (as consolidation temporarily acquires an exclusive lock on the array).
```
teo-tsirpanis, line 252:
```suggestion
// begin (as consolidation temporarily acquires an exclusive lock on the array).
// Deprecated: Use ConsolidateArray instead.
```

array.go
```diff
@@ -208,59 +208,76 @@ func (a *Array) Close() error {
...
 	if ret != C.TILEDB_OK {
-		return fmt.Errorf("error vacuuming tiledb array: %w", a.context.LastError())
+		return fmt.Errorf("error vacuuming tiledb array: %w", tdbCtx.LastError())
 	}
 
 	runtime.KeepAlive(config)
 	return nil
 }
 
+// Vacuum cleans up the array, such as consolidated fragments and array metadata.
```
teo-tsirpanis, line 276:
```suggestion
// Vacuum cleans up the array, such as consolidated fragments and array metadata.
// Deprecated: Use VacuumArray instead.
```

shaunrd0 (commented):
Looks like there are conflicts from #370 but LGTM once those are resolved, thanks for doing this!

array.go
teo-tsirpanis: > I don't see a reason to deprecate?

The reason is the same with the `Create` APIs: there is no C API to consolidate/vacuum an already opened array. TileDB-Go so far has provided a mostly 1:1 API surface with the C API, and these methods require extra book-keeping of the URI that I would like to remove.

> a client would probably just prefix with calls to Array.URI() and Array.Context() anyhow

This might be a valid scenario for consolidation, but people really shouldn't vacuum an already open array. 😅

shaunrd0: Ah I see, thanks for the clarification! SGTM if you'd like to deprecate them

group.go
```diff
@@ -36,19 +36,24 @@ func NewGroup(tdbCtx *Context, uri string) (*Group, error) {
...
+	ret := C.tiledb_group_create(tdbCtx.tiledbContext, curi)
+	runtime.KeepAlive(tdbCtx)
 	if ret != C.TILEDB_OK {
-		return fmt.Errorf("error in creating group: %w", g.context.LastError())
+		return fmt.Errorf("error in creating group: %w", tdbCtx.LastError())
 	}
 	return nil
 }
 
+// Create creates a new TileDB group.
```
teo-tsirpanis, line 52:
```suggestion
// Create creates a new TileDB group.
// Deprecated: Use CreateGroup instead.
```

group.go
teo-tsirpanis: Added suggestion in the correct place. Thanks for noticing!

array.go
```diff
@@ -208,59 +208,76 @@ func (a *Array) Close() error {
...
-func (a *Array) Vacuum(config *Config) error {
+// Consolidate consolidates the fragments of the array into a single fragment.
+// You must first finalize all queries to the array before consolidation can
+// begin (as consolidation temporarily acquires an exclusive lock on the array).
+func (a *Array) Consolidate(config *Config) error {
+	return ConsolidateArray(a.context, a.uri, config)
+}
+
+// VacuumArray cleans up an array, such as consolidated fragments and array metadata.
+func VacuumArray(tdbCtx *Context, uri string, config *Config) error {
```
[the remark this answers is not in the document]
teo-tsirpanis: Done.

array.go
```diff
@@ -208,59 +208,76 @@ func (a *Array) Close() error {
 	return nil
 }
 
-// Create creates a new TileDB array given an input schema.
-func (a *Array) Create(arraySchema *ArraySchema) error {
-	curi := C.CString(a.uri)
+// Create creates a new TileDB array given a context, URI and schema.
+func CreateArray(tdbCtx *Context, uri string, arraySchema *ArraySchema) error {
```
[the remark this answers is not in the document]
teo-tsirpanis: Fixed comment.

PR #369, opened 2025-03-10
Update to golang 1.24.

shaunrd0:
This updates to golang 1.24 and converts usages of `runtime.SetFinalizer` to `runtime.AddCleanup`.

### DIFF
--- a/.github/workflows/tiledb-go.yml
+++ b/.github/workflows/tiledb-go.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         # Will be checking following versions
-        go: ["1.24", "1.25"]
+        go: ["1.23", "1.24"]
     steps:
 
     # Checks out repository
@@ -74,7 +74,7 @@ jobs:
     strategy:
       matrix:
         # Will be checking following versions
-        go: ["1.24", "1.25"]
+        go: ["1.23", "1.24"]
     steps:
     # Checks out repository
     - uses: actions/checkout@v4
@@ -113,7 +113,7 @@ jobs:
     strategy:
       matrix:
         # Will be checking following versions
-        go: ["1.24", "1.25"]
+        go: ["1.23", "1.24"]
     steps:
     # Checks out repository
     - uses: actions/checkout@v4
--- a/.github/workflows/tiledb-go.yml
+++ b/.github/workflows/tiledb-go.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         # Will be checking following versions
-        go: ["1.22", "1.23"]
+        go: ["1.24", "1.25"]
     steps:
 
     # Checks out repository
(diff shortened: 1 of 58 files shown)


.github/workflows/tiledb-go.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     strategy:
       matrix:
         # Will be checking following versions
-        go: ["1.22", "1.23"]
+        go: ["1.23", "1.24"]
```
shaunrd0, line 40: Golang 1.25 is not released yet, so I stepped back to 1.23 instead testing 1.24 alone. If we don't need to look backwards like this I can remove and just test against 1.24.

array.go
```diff
@@ -71,7 +71,7 @@ func NewArray(tdbCtx *Context, uri string) (*Array, error) {
 	if ret != C.TILEDB_OK {
 		return nil, fmt.Errorf("error creating tiledb array: %w", array.context.LastError())
 	}
-	freeOnGC(&array)
+	runtime.AddCleanup(&array, freeFreeable, Freeable(&array))
```
teo-tsirpanis, line 74: I've been thinking of another approach that will avoid the `runtime.AddCleanup` boilerplate everywhere and maybe support both old and new-style finalizers.

---

Also we should keep the `Cleanup` object somewhere so that we remove it if the object is manually `Free()`ed.

examples/array_metadata_test.go
```diff
@@ -42,9 +42,9 @@ import (
 	"github.com/TileDB-Inc/TileDB-Go/examples_lib"
 )
 
-// ExampleArrayMetadataArray shows and example creation, writing and reading of a
-// sparse array
-func ExampleArrayMetadataArray() {
+// ExampleRunArrayMetadataArray shows an example of creation, writing, and
+// reading of a sparse array
+func ExampleRunArrayMetadataArray() {
```
teo-tsirpanis, line 47: The changes to the examples look unrelated; can you remove them from the PR?

In a subsequent PR we should fold the `examples_lib` files into `examples`.

.github/workflows/tiledb-go.yml
teo-tsirpanis: This can't actually run on Go 1.23 because of the version declaration in `go.mod`. I guess CI still runs it in 1.24?

teo-tsirpanis (changes requested):

array.go
shaunrd0: Can you create a story? I can revert for now, and we can sync on this tomorrow morning or you can pick it up if you'd like :slightly_smiling_face:

teo-tsirpanis: Opened SC-64250.

examples/array_metadata_test.go
teo-tsirpanis: Yeah, I tried directly calling the functions in `examples` but couldn't for no apparent reason. Such a shame.

shaunrd0:
Closing after syncing with @teo-tsirpanis, the upgrade to golang 1.24 will be done in https://github.com/TileDB-Inc/TileDB-Go/pull/370

PR #370, opened 2025-03-11
Refactor handle management.

teo-tsirpanis:
[SC-64250](https://app.shortcut.com/tiledb-inc/story/64250)

This PR refactors the intrastructure managing the lifetime of native C API handles. The goals are the following:

1. Use Go 1.24's improved finalizer API.
2. Deregister the finalizer if the user manually calls `Free` on an object.
3. Improve safety.
4. Minimize the amount of added boilerplate.
5. Maintain compatibility with previous Go versions.

We introduce a new type, `capiHandle` which is responsible for managing the lifetime of the native handles. When creating an object, we now register the finalizer on the `*capiHandle` instead of the returned pointer. By the use of this extra level of indirection and atomic operations, we prevent double-frees and use-after-frees, including if you accidentally copy an object like we do in https://github.com/TileDB-Inc/TileDB-Go/pull/366#discussion_r1956703094. To provide type safety, we introduce specific handle types that embed a `*capiHandle`.

There are two mostly identical implementations of `capiHandle`; for Go versions before and since 1.24. Once Go 1.23 goes out of support, we should remove the downlevel implemenation.

### DIFF
--- a/array.go
+++ b/array.go
@@ -23,7 +23,7 @@ e.g. on disk, in an S3 bucket, etc. Once an array has been opened for reading
 or writing, interact with the data through Query objects.
 */
 type Array struct {
-	tiledbArray *C.tiledb_array_t
+	tiledbArray *capiHandle[C.tiledb_array_t]
 	context     *Context
 	uri         string
 	config      *Config
@@ -54,6 +54,10 @@ type NonEmptyDomain struct {
 	Bounds        interface{}
 }
 
+func newArrayFromHandle(tdbCtx *Context, arrayHandle *capiHandle[C.tiledb_array_t]) *Array {
+	return &Array{context: tdbCtx, tiledbArray: arrayHandle}
+}
+
 // NewArray allocates a new array.
 // If the provided Context is nil, a default context is allocated and used.
 func NewArray(tdbCtx *Context, uri string) (*Array, error) {
@@ -66,14 +70,15 @@ func NewArray(tdbCtx *Context, uri string) (*Array, error) {
 	}
 	curi := C.CString(uri)
 	defer C.free(unsafe.Pointer(curi))
-	array := Array{context: tdbCtx, uri: uri}
-	ret := C.tiledb_array_alloc(array.context.tiledbContext, curi, &array.tiledbArray)
+	var arrayPtr *C.tiledb_array_t
+	ret := C.tiledb_array_alloc(tdbCtx.tiledbContext, curi, &arrayPtr)
 	if ret != C.TILEDB_OK {
-		return nil, fmt.Errorf("error creating tiledb array: %w", array.context.LastError())
+		return nil, fmt.Errorf("error creating tiledb array: %w", tdbCtx.LastError())
 	}
-	freeOnGC(&array)
+	array := newArrayFromHandle(tdbCtx, newArrayHandle(arrayPtr))
+	array.uri = uri
 
-	return &array, nil
+	return array, nil
 }
 
 // Free releases the internal TileDB core data that was allocated on the C heap.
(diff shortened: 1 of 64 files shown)


teo-tsirpanis:
@TileDB-Inc/tiledb-cloud-core could you take a first look at my approach, before I port the rest of the codebase?

I ported three handles; one simple case (array) and two with some additional considerations (array schema, where the handle of an object can be reassigned and buffer, where we have a pinner that we must release after the handle).

capi_handle_legacy.go
```diff
@@ -0,0 +1,55 @@
+//go:build !go1.24
```
shaunrd0, line 1: :+1:

array_schema.go
```diff
@@ -11,9 +11,54 @@ import (
...
+	return newCapiHandle(unsafe.Pointer(ptr), freeCapiArraySchema)
+}
+
+func newArraySchemaHandle(ptr *C.tiledb_array_schema_t) arraySchemaHandle {
+	h := newArraySchemaCapiHandle(ptr)
+	return arraySchemaHandle{h: unsafe.Pointer(h)}
+}
+
+func (x *arraySchemaHandle) loadCapiHandle() *capiHandle {
+	return (*capiHandle)(atomic.LoadPointer(&x.h))
```
shaunrd0, line 42: `loadCapiHandle` and `Free` don't seem to need the arraySchemaHandle type, could they be defined in some base interface so that we don't duplicate those lines for every TileDB-Go type we want to manage? 

Kinda ditto for my last comment I guess, it seems like we could use an interface with some default implementation to save some code duplication redefining these.

array_schema.go
```diff
@@ -11,9 +11,54 @@ import (
...
 )
 
+type arraySchemaHandle struct {
+	// An unsafe pointer to the capiHandle holding the schema.
+	// This pointer must be accessed using atomic functions, because the MarshalJSON function
+	// replaces the native handle of an existing object.
+	// We cannot wrap atomic.Pointer[capiHandle] in a struct and pass it around because of lint warnings,
+	// and we cannot mutate the native handle of an existing capiHandle because it's impossible to atomically
+	// set the finalizer.
+	h unsafe.Pointer
```
shaunrd0, line 25: Could this struct hold any capiHandle since it's only member is an `unsafe.Pointer`?  I'm noticing it seems not specific to the array schema aside from definitions for `Get` and `Reset`, and it feels like other TileDB-Go types would do something similar and wrap `capiHandle` with `unsafe.Pointer`?

shaunrd0 (commented):
Looks great to me so far, nice work! I left some comments to think about but nothing stands out as a concern.

array_schema.go
teo-tsirpanis: `arraySchemaHandle` is special because the `*capiHandle` can be mutated, which means that all accesses to `h` must be atomic. The other handles are immutable and don't need the extra level of indirection in `loadCapiHandle`, they can just read it.

array_schema.go
shaunrd0: No worries, feel free to disregard - I'm probably thinking too much like C++. If ArraySchemaHandle is the only type that needs to use atomics as you mentioned in your reply to my other comment then there's really no point to thinking about a base class anyhow.

array.go
```diff
@@ -66,14 +81,15 @@ func NewArray(tdbCtx *Context, uri string) (*Array, error) {
...
-	ret := C.tiledb_array_alloc(array.context.tiledbContext, curi, &array.tiledbArray)
+	var arrayPtr *C.tiledb_array_t
+	ret := C.tiledb_array_alloc(tdbCtx.tiledbContext.Get(), curi, &arrayPtr)
 	if ret != C.TILEDB_OK {
-		return nil, fmt.Errorf("error creating tiledb array: %w", array.context.LastError())
+		return nil, fmt.Errorf("error creating tiledb array: %w", tdbCtx.LastError())
 	}
-	freeOnGC(&array)
+	array := newArrayFromHandle(tdbCtx, newArrayHandle(arrayPtr))
+	array.uri = uri
```
shaunrd0, line 90: Should the URI be set in `newArrayFromHandle` so that an Array created from handles in other places have the URI initialized?

teo-tsirpanis: `NewArray` is the only case where we've been setting the `uri` field, which is used only by some APIs that should be deprecated either way (#367).

PR #371, opened 2025-04-29
Update documentation links

nickvigilante:
Also removes some links pointing to docs.tiledb.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,7 @@ free'ing of C heap allocated structures.
 ## Quick Links
 
 - GoDoc API documentation: [https://pkg.go.dev/github.com/TileDB-Inc/TileDB-Go](https://pkg.go.dev/github.com/TileDB-Inc/TileDB-Go)
-- Full Installation Docs: [https://docs.tiledb.com/main/how-to/installation](https://docs.tiledb.com/main/how-to/installation)
-- Quick Install: [https://docs.tiledb.com/main/how-to/installation/quick-install](https://docs.tiledb.com/main/how-to/installation/quick-install)
-- Full developer documentation for all APIs and integrations: [https://docs.tiledb.com](https://docs.tiledb.com)
+- Full developer documentation for all APIs and integrations: [https://cloud.tiledb.com/academy](https://cloud.tiledb.com/academy)
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


teo-tsirpanis (changes requested):
This document has been kind of a misnomer because building Go packages from source is not really a thing. It is also duplicated by the README.

nickvigilante:
@teo-tsirpanis Maybe it would be best then to close this PR, since it's duplicated content? I didn't read through the README thoroughly enough.

README.md
```diff
@@ -12,9 +12,8 @@ free'ing of C heap allocated structures.
 ## Quick Links
 
 - GoDoc API documentation: [https://pkg.go.dev/github.com/TileDB-Inc/TileDB-Go](https://pkg.go.dev/github.com/TileDB-Inc/TileDB-Go)
-- Full Installation Docs: [https://docs.tiledb.com/main/how-to/installation](https://docs.tiledb.com/main/how-to/installation)
-- Quick Install: [https://docs.tiledb.com/main/how-to/installation/quick-install](https://docs.tiledb.com/main/how-to/installation/quick-install)
-- Full developer documentation for all APIs and integrations: [https://docs.tiledb.com](https://docs.tiledb.com)
+- Building from Source: [docs/building-from-source.md](docs/building-from-source.md)
```
shaunrd0, line 15: I'd also remove this line since the information is in the README as Theo points out. The surrounding changes LGTM, the other links removed are stale and I think adding a link to the academy is a good change

```suggestion

```

README.md
```diff
@@ -27,7 +26,7 @@ Currently the following platforms are supported:
 
 ### Prerequisites
 This package requires the TileDB shared library be installed and on the system path. See the
-[official TileDB installation instructions](https://docs.tiledb.com/main/how-to/installation)
+[official TileDB installation instructions](docs/building-from-source.md)
```
shaunrd0, line 29: This is for TileDB core, so I think we can just link to the docs recently added there

```suggestion
[TileDB GitHub repository](https://github.com/TileDB-Inc/TileDB/blob/main/BUILDING_FROM_SOURCE.md)
```

shaunrd0 (commented):
Thanks @nickvigilante! I agree with @teo-tsirpanis. We could still take this opportunity to update the links in the README but I think the `docs/building-from-source.md` can be removed.

teo-tsirpanis (approved):
Thanks!

PR #375, opened 2025-06-23
add group metadata functions with similar pattern that arrays use

disideris:
This PR fixes [ENG-98](https://linear.app/tiledb/issue/ENG-98/add-support-for-group-metadata-in-tiledb-go).

While working with asset metadata and specifically group metadata i noticed that we could add some functions in `group.go` so that we can have a similar pattern on how to interact with core metadata with arrays and groups. All the infra and core support is there we were just missing the functions from this PR.

### DIFF
--- a/group.go
+++ b/group.go
@@ -35,6 +35,11 @@ type Group struct {
 	context *Context
 }
 
+// Context exposes the internal TileDB context used to initialize the group.
+func (g *Group) Context() *Context {
+	return g.context
+}
+
 func newGroupFromHandle(context *Context, uri string, group groupHandle) *Group {
 	return &Group{group: group, uri: uri, context: context}
 }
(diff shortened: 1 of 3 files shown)


group.go
```diff
@@ -349,6 +354,39 @@ func (g *Group) GetMetadata(key string) (Datatype, uint, interface{}, error) {
 	return datatype, valueNum, value, nil
 }
 
+// GetMetadataMap returns a map[string]*GroupMetadata where key is the key of
+// each metadata added and value is an GroupMetadata struct. The map contains
+// all group metadata previously added.
```
teo-tsirpanis, line 359:
```suggestion
// GetMetadataMap returns a map with the group's metadata, indexed by their key.
```

group.go
```diff
@@ -35,6 +35,11 @@ type Group struct {
 	context *Context
 }
 
+// Context exposes the internal TileDB context used to initialize the group.
```
teo-tsirpanis, line 38:
```suggestion
// Context returns the TileDB context used to initialize the group.
```

group.go
```diff
@@ -349,6 +354,39 @@ func (g *Group) GetMetadata(key string) (Datatype, uint, interface{}, error) {
...
+func (g *Group) GetMetadataMap() (map[string]*GroupMetadata, error) {
+	return g.GetMetadataMapWithValueLimit(nil)
+}
+
+// GetMetadataMapWithValueLimit returns a map[string]*GroupMetadata where key is the key of
+// each metadata added and value is an GroupMetadata struct. The map contains
+// all group metadata previously added.
+// The limit parameter limits the number of values returned if string or group.
+// This is helpful for pushdown of limiting metadata. If nil, value is returned
+// in full.
```
teo-tsirpanis, line 369:
```suggestion
// GetMetadataMapWithValueLimit returns a map with the group's metadata, indexed by their key.
// The limit parameter limits the size of values returned if string or array.
// This is helpful for pushdown of limiting metadata. If nil, value is returned
// in full.
```

group.go
```diff
@@ -349,6 +354,39 @@ func (g *Group) GetMetadata(key string) (Datatype, uint, interface{}, error) {
...
+func (g *Group) GetMetadataMapWithValueLimit(limit *uint) (map[string]*GroupMetadata, error) {
```
ypatia, line 370: Could we add tests for those?

teo-tsirpanis, line 370: I don't like that `limit` is an unsigned integer but I'm fine with it to maintain consistency with `Array`.

teo-tsirpanis (approved):
LGTM, thanks! Documentation can be improved (please also update the corresponding functions on `Array`).

PR #378, opened 2025-07-16
Use `*_dump_str` instead of deleted `*_dump`

kounelisagis:


attribute.go
```diff
@@ -388,38 +388,50 @@ func (a *Attribute) Type() (Datatype, error) {
...
 	runtime.KeepAlive(a)
 	if ret != C.TILEDB_OK {
-		return fmt.Errorf("error dumping attribute to stdout: %w", a.context.LastError())
+		return fmt.Errorf("error dumping attribute to string: %w", a.context.LastError())
 	}
+	defer C.tiledb_string_free(&cStr)
+
+	var cStrPtr *C.char
+	var cStrLen C.size_t
+	ret = C.tiledb_string_view(cStr, &cStrPtr, &cStrLen)
```
teo-tsirpanis, line 401: You should use the `stringHandleToString` helper function.

teo-tsirpanis (commented):
We should add new `DumpToString()` functions that return a string ([example](https://github.com/TileDB-Inc/TileDB-Go/blob/8452081fbb8754fee008a07e33ef2fe85fcccf45/group.go#L455-L484)), and  implement the existing dump functions in terms of these.

teo-tsirpanis (approved):
Thanks!

